PR #32, opened 2015-11-16
Adds google geocoder

gfiorav:
Please CR, @ethervoid 

As soon as we can connect REDIS to get the geocoder we need to use, this should work right off the bat.

cc/ @rafatower


server/lib/python/googlemapswrapper/googlemapswrapper/googlemapsgeocoder.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/local/bin/python
+# -*- coding: utf-8 -*-
+
+import googlemaps
```
rafatower: @ethervoid Just to mark down in our notebooks, this is a dependency to be installed. I guess this is the source:

https://github.com/googlemaps/google-maps-services-python

PR #77, opened 2016-02-02
Commit cdb_geocoder_server--0.1.0.sql into VC

ethervoid:
We need to store this files because they are used in the execution
of the make file and if one of this doesn't exists the make execution
is going to crash like this:

make: *** No rule to make target `cdb_geocoder_server--0.1.0.sql',
needed by `all'.  Stop.

This commit also fixes the problem with the CI tests


rafatower:
I just made sure it's the same file deployed to servers:
```
[rtorre@rtorre-cartodb geocoder-api (store_master_sql_file)]$ md5sum server/extension/cdb_geocoder_server--0.1.0.sql
0cc5a4aeaa9df12611909df0aad493a8  server/extension/cdb_geocoder_server--0.1.0.sql
```

```
$ md5sum /usr/share/postgresql/9.3/extension/cdb_geocoder_server--0.1.0.sql
0cc5a4aeaa9df12611909df0aad493a8  /usr/share/postgresql/9.3/extension/cdb_geocoder_server--0.1.0.sql
```

For the same reason, the migration file shall also be committed (making sure it is the same deployed to the servers).

I have mixed feelings about whether they should be in `.gitignore` or not (mind the 0.0.1 right up) . We shall also tag it properly after adding those files to master.

ethervoid:
retest this please

PR #85, opened 2016-02-11
Routing isolines functions

ethervoid:
- Trade areas using HERE maps routing API


client/cdb_dataservices_client--0.1.0--0.2.0.sql
```diff
@@ -0,0 +1,60 @@
+CREATE TYPE cdb_dataservices_client.isoline AS (
+    center geometry(Geometry,4326),
+    data_range integer,
+    the_geom geometry(Multipolygon,4326)
+);
+
+CREATE OR REPLACE FUNCTION cdb_dataservices_client.cdb_isodistance (source geometry(Geometry, 4326), mode text, range integer[], options text[] DEFAULT NULL)
+RETURNS SETOF cdb_dataservices_client.isoline AS $$
```
rafatower: :+1:

client/renderer/interfaces/interface_0.2.0.yaml
```diff
@@ -0,0 +1,79 @@
...
+  return_type: SETOF cdb_dataservices_client.isoline
+  multi_row: true
+  params:
+    - { name: source,  type: "geometry(Geometry, 4326)" }
+    - { name: mode, type: text }
+    - { name: range, type: "integer[]" }
+    - { name: options, type: "text[]", default: 'NULL' }
+
+- name: cdb_isochrone
+  return_type: SETOF cdb_dataservices_client.isoline
```
rafatower: :+1: 

I guess we can release (now) this version of the client but then it'll be a bit behind the server. We'd need to start thinking on taking into account client changes and versions for reference documentation.

So far what I saw is just use latest.

client/sql/0.2.0/15_config_management.sql
```diff
@@ -0,0 +1,37 @@
+CREATE TYPE cdb_dataservices_client._entity_config AS (
```
rafatower: For all those files, when there are no changes we could use symlinks.

This would help a lot in reviews and clarity of the code.

server/extension/cdb_dataservices_server--0.3.0--0.4.0.sql
```diff
@@ -0,0 +1,115 @@
...
+      quota_service.increment_empty_geocoder_use()
+  except BaseException as e:
+    import sys, traceback
+    type_, value_, traceback_ = sys.exc_info()
+    quota_service.increment_failed_geocoder_use()
+    error_msg = 'There was an error trying to obtain isodistances using here maps geocoder: {0}'.format(e)
+    plpy.notice(traceback.format_tb(traceback_))
+    plpy.error(error_msg)
+  finally:
+    quota_service.increment_total_geocoder_use()
```
rafatower: geocoder use??

this is to be changed, right?

server/extension/sql/0.4.0/10_redis_helper.sql
```diff
@@ -0,0 +1,57 @@
...
+        metrics_config_params['sentinel_port'],
+        metrics_config_params['sentinel_master_id'],
+        timeout=metrics_config_params['timeout'],
+        redis_db=metrics_config_params['redis_db']).redis_connection()
+    GD[cache_key] = {
+      'redis_metadata_connection': redis_metadata_connection,
+      'redis_metrics_connection': redis_metrics_connection,
+    }
+    return True
+$$ LANGUAGE plpythonu SECURITY DEFINER;
```
rafatower: (please let me know if there are changes worth checking in these helpers between 0.3.0 and 0.4.0

rafatower:
:+1: please mind this is hard to review so guide me if I might be missing something

client/sql/0.2.0/15_config_management.sql
ethervoid: That could be a pain becuase we have to reference to other commits or something like that?

server/extension/cdb_dataservices_server--0.3.0--0.4.0.sql
ethervoid: Yes, that is what I commented. After this release, we need to release a minor version in the server to solve some naming issues like this.

server/extension/sql/0.4.0/10_redis_helper.sql
ethervoid: All the changes are reflected in the upgrade and downgrade files.

PR #87, opened 2016-02-15
New data services api docs

iriberri:
Here are the changes of the doc to rename it as Data Services API and adding the new routing functions.

I have separated the previous functions section into two, per each service.

@ethervoid worked on the routing functions details and I integrated the whole thing together.

Please, **don't merge** as changes will need to be done in the `docs` repo, but this PR can be used for reviewing.

cc @csobier could you please give this a look? :-)


iriberri:
cc @rafatower

ethervoid:
:clap:

doc/API.md
```diff
@@ -1,9 +1,14 @@
-# Geocoder API
+# Data Services API
 
-The CartoDB Geocoder API allows you to match your data with geometries on your map. This geocoding service can be used programatically to geocode datasets via the CartoDB SQL API. It is fed from _Open Data_ and it serves geometries for countries, provinces, states, cities, postal codes, IP addresses and street addresses.
+The CartoDB Data Services API offers a set of location based services that can be used programatically to empower your geospatial applications.
+
+The Geocoder functions allow you to match your data with geometries on your map. This geocoding service can be used programatically to geocode datasets via the CartoDB SQL API. It is fed from _Open Data_ and it serves geometries for countries, provinces, states, cities, postal codes, IP addresses and street addresses.
+
+The Routing functions provide a way to generate isolines in terms of distance and time by means of the available isodistance and isochrone functions useful for Trade Areas Analysis.
```
rafatower: I'm not strongly opposed but probably I wouldn't mention routing here as it can be confusing. It is incidental finding isolines within HERE routing API.

doc/general_concepts.md
```diff
@@ -1,20 +1,20 @@
 # General concepts
 
-The Geocoder API offers geocoding services on top of the CartoDB SQL API by means of a set of geocoding functions. Each one of these functions is oriented to one kind of geocoding operation and it will return the corresponding geometry (a `polygon` or a `point`) according to the input information.
+The Data Services API offers geocoding and routing services on top of the CartoDB SQL API by means of a set of functions. Each one of these functions is oriented to one kind of operation and it will return the corresponding geometry (a `polygon` or a `point`) according to the input information.
 
-The Geocoder API decouples the geocoding service from the CartoDB Editor, and allows to geocode data (being single rows, complete datasets or simple inputs) programatically through authenticated requests.
+The Data Services API decouples the geocoding and routing services from the CartoDB Editor. The API allows to geocode data (being single rows, complete datasets or simple inputs) and to perform routing analysis (computing isodistances or isochrones) programatically through authenticated requests.
```
rafatower: > and to perform routing analysis (computing isodistances or isochrones) [...]

and perform trade areas analysis (computing isodistances or isochrones) [...]

doc/routing_functions.md
```diff
@@ -0,0 +1,92 @@
...
+SELECT the_geom FROM cdb_isochrone('POINT(-3.70568 40.42028)'::geometry, 'walk', ARRAY[300]::integer[]);
+```
+
+### Optional isoline parameters
+
+The optional value parameters must be passed with the format: `option=value`.
+
+Name | Type | Description | Accepted values
+--- | --- | --- | ---
+`is_destination` | `boolean` | If true, the source point is the destination instead of the starting location | `true` or `false`
```
rafatower: missing the default

doc/routing_functions.md
```diff
@@ -0,0 +1,92 @@
...
+`mode_type` | `text` | Type of route calculation | `shortest` or `fastest`. `shortest` by default
+`mode_traffic` | `text` | Use traffic data to calculate the route | `enabled` or `disabled`. `disabled` by default
+`singlecomponent` | `boolean` | If true, the isoline service will return a single polygon area, instead of creating a separate polygon for each reachable area separated of the main polygon (known as island). | `true` or `false`. `false` by default
```
rafatower: I'd remove this one (unnecessary complexity) at least from the docs and keep it as `true` by default.

Otherwise it is calling for problems. The original intent of this param in HERE side is to offer geometries connected somehow when for instance transport is ferry... something that leaves us more exposed to lack of regularity in return params from HERE.

iriberri: Makes sense :+1:

@ethervoid this would imply setting this option always in the Python functionality, right?

ethervoid: Yes we should change it to use `true` by default and remove that entry from the docs

csobier:
I applied some tiny, minor grammar fixes. That's it.  Looks good! :+1:

iriberri:
@ethervoid @rafatower I just removed "routing" from everywhere ;)

iriberri:
we also need to make clear that it's a paid service, I'll do it asap.

iriberri:
... and know I'm thinking about moving the bullet list about quotas to an external file and redirect to it from each section in which money is involved, instead of having the info duplicated twice (and so on when we add more services).

rafatower:
Related to https://github.com/CartoDB/docs/pull/692

rafatower:
Watch out for conflicts, worth updating with master and passing the tests cc/ @ethervoid 

@iriberri: since the whole doc update involves 2 repositories and 3 PR's I'd like you to take a final look when everything is in place (just "sanity testing" the final result)

Thank you all, good job! :+1:

ethervoid:
@iriberri I have solved the conflicts, check out it's all ok

csobier:
As per @rafatower , merging.

PR #90, opened 2016-02-17
Support using a common redis connection

luisbosque:
If sentinel_host or sentinel_cluster_id is not provided it will try to
connect with a redis_host parameter

This way we can support deploying simpler local servers environments like a development one.

Redis command can be run in the same way from a strictredis connection instance or a sentinel one through master_for method, so the type of connection returned by python should be totally transparent for the extension.

It's a proof of concept. If it's ok I should probably add a test for this new case and we should decide in what case a strictredis connection is chosen over of sentinel one. At this moment strictredis connection is chosen when sentinel_host or sentinel_cluster_id are None. Either if we choose to introduce this support or not it would be a good a idea to add some params checking in the python lib connection method.

/cc @rafatower @ethervoid

### DIFF
--- a/server/lib/python/cartodb_services/cartodb_services/tools/redis_tools.py
+++ b/server/lib/python/cartodb_services/cartodb_services/tools/redis_tools.py
@@ -6,28 +6,26 @@ class RedisConnection:
 
     REDIS_DEFAULT_USER_DB = 5
     REDIS_DEFAULT_TIMEOUT = 2  #seconds
-    REDIS_SENTINEL_DEFAULT_PORT = 26379
-    REDIS_DEFAULT_PORT = 6379
+    #REDIS_SENTINEL_DEFAULT_PORT = 26379
+    #REDIS_DEFAULT_PORT = 6379
 
-    def __init__(self, sentinel_host, sentinel_port, sentinel_master_id,
-                 redis_host, redis_db=REDIS_DEFAULT_USER_DB, **kwargs):
-        self.sentinel_host = sentinel_host
-        self.sentinel_port = sentinel_port
-        self.sentinel_master_id = sentinel_master_id
+    def __init__(self, sentinel_master_id, redis_host, redis_port,
+                 redis_db=REDIS_DEFAULT_USER_DB, **kwargs):
         self.redis_host = redis_host
+        self.redis_port = redis_port
+        self.sentinel_master_id = sentinel_master_id
         self.timeout = kwargs['timeout'] if 'timeout' in kwargs else self.REDIS_DEFAULT_TIMEOUT
         self.redis_db = redis_db
-        self.redis_port = self.REDIS_DEFAULT_PORT
 
     def redis_connection(self):
         return self.__create_redis_connection()
 
     def __create_redis_connection(self):
-        if (self.sentinel_host == None) or (self.sentinel_master_id == None):
+        if self.sentinel_master_id == None:
             return StrictRedis(host=self.redis_host, port=self.redis_port, db=self.redis_db)
         else:
-            sentinel = Sentinel([(self.sentinel_host,
-                                  self.REDIS_SENTINEL_DEFAULT_PORT)],
+            sentinel = Sentinel([(self.redis_host,
+                                  self.redis_port)],
                                 socket_timeout=self.timeout)
             return sentinel.master_for(
                 self.sentinel_master_id,
--- a/server/lib/python/cartodb_services/cartodb_services/tools/redis_tools.py
+++ b/server/lib/python/cartodb_services/cartodb_services/tools/redis_tools.py
@@ -1,4 +1,5 @@
 from redis.sentinel import Sentinel
+from redis import StrictRedis
 
 
 class RedisConnection:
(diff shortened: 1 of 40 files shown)


rafatower:
@ethervoid if you're ok with the changes we should think on merging these with the version you're working on

ethervoid:
No time to add in this deploy, I will check it in the next deploy

luisbosque:
No urgent at all. Are you ok with this strategy then?

rafatower:
I'm OK with supporting both sentinel and direct connections. This PR
uncovers issues with versioning and branchy development from our side...

server/lib/python/cartodb_services/cartodb_services/tools/redis_tools.py
```diff
@@ -6,24 +7,30 @@ class RedisConnection:
...
     def __create_redis_connection(self):
-        sentinel = Sentinel([(self.sentinel_host,
-                              self.REDIS_SENTINEL_DEFAULT_PORT)],
-                            socket_timeout=self.timeout)
-        return sentinel.master_for(
-            self.sentinel_master_id,
-            socket_timeout=self.timeout,
-            db=self.redis_db
-        )
+        if (self.sentinel_host == None) or (self.sentinel_master_id == None):
```
ethervoid: Here we could catch the sentinel exception if the master id doesn't return a master instance and use the redis direct connection

luisbosque: Uhm, that make perfects sense.

You mean having just one redis_host and redis_port in the conf and rely on
the exception error for the type of redis guessing, right?

On Fri, Feb 19, 2016 at 11:01 AM, Mario de Frutos <notifications@github.com>
wrote:

> In
> server/lib/python/cartodb_services/cartodb_services/tools/redis_tools.py
> <https://github.com/CartoDB/dataservices-api/pull/90#discussion_r53441947>
> :
>
> >
> >      def redis_connection(self):
> >          return self.__create_redis_connection()
> >
> >      def __create_redis_connection(self):
> > -        sentinel = Sentinel([(self.sentinel_host,
> > -                              self.REDIS_SENTINEL_DEFAULT_PORT)],
> > -                            socket_timeout=self.timeout)
> > -        return sentinel.master_for(
> > -            self.sentinel_master_id,
> > -            socket_timeout=self.timeout,
> > -            db=self.redis_db
> > -        )
> > +        if (self.sentinel_host == None) or (self.sentinel_master_id == None):
>
> Here we could catch the sentinel exception if the master id doesn't return
> a master instance and use the redis direct connection
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/dataservices-api/pull/90/files#r53441947>.
>



-- 
Luis Bosque
@luisico

www.cartodb.com
Map, analyze and build applications with your data

luisbosque: The only "but" I see on that is that testing the redis connection would be a little bit messy, right?

rafatower: Yeah, It seems better the "configuration approach" vs the "probe and
fallback approach".

On Fri, Feb 19, 2016 at 11:57 AM, luisico <notifications@github.com> wrote:

> In
> server/lib/python/cartodb_services/cartodb_services/tools/redis_tools.py
> <https://github.com/CartoDB/dataservices-api/pull/90#discussion_r53447447>
> :
>
> >
> >      def redis_connection(self):
> >          return self.__create_redis_connection()
> >
> >      def __create_redis_connection(self):
> > -        sentinel = Sentinel([(self.sentinel_host,
> > -                              self.REDIS_SENTINEL_DEFAULT_PORT)],
> > -                            socket_timeout=self.timeout)
> > -        return sentinel.master_for(
> > -            self.sentinel_master_id,
> > -            socket_timeout=self.timeout,
> > -            db=self.redis_db
> > -        )
> > +        if (self.sentinel_host == None) or (self.sentinel_master_id == None):
>
> The only "but" I see on that is that testing the redis connection would be
> a little bit messy, right?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/dataservices-api/pull/90/files#r53447447>.
>



-- 
*Rafa de la Torre*
DataServices Engineer www.cartodb.com

luisbosque: In that case, in order to make configuration much easier we can have this conf:
* redis_host: string
* redis_port: int
* use_sentinel: bool

That way we don't bother in doing "guessing". Just follow the value in use_sentinel.
What do you think?

ethervoid: I like that proposal :)

luisbosque: Or even better, in order to avoid adding another parameter we can connect with sentinel if there's sentinel_master_id and with redis if not.

luisbosque:
retest this please

ethervoid:
Merged in this [PR](https://github.com/CartoDB/dataservices-api/pull/91)

PR #91, opened 2016-02-22
Mapzen integration

### LINKED ISSUE #92: remove sentinel dependency
in order to configure it locally sentinel is needed, this should be needed (and this configuration should not be there actually)

ethervoid:
**We have to merge PR with the Redis config changes in Chef**

Closes #92 

- Added redis direct connection option if there is no `master_id` key defined (thx to @luisbosque)
- Mapzen turn-by-turn integration (in its simplest form, from point A to point B)


ethervoid:
@rafatower check this PR please

server/extension/cdb_dataservices_server--0.4.0--0.5.0.sql
```diff
@@ -0,0 +1,157 @@
...
+        "redis_db": params['redis_db']
+      }
+      if "sentinel_master_id" in params:
+        redis_conf_params["sentinel_master_id"] = params["sentinel_master_id"]
+      else:
+        redis_conf_params["sentinel_master_id"] = None
+
+      return redis_conf_params
+$$ LANGUAGE plpythonu;
+
```
rafatower: luisico's patch merged here? :smile:

server/extension/sql/0.5.0/100_routing_helper.sql
```diff
@@ -0,0 +1,55 @@
+CREATE TYPE cdb_dataservices_server.simple_route AS (
```
rafatower: minor detail: this file is called "helper" when indeed it implements the core of the routing.

rafatower:
:+1:

PR #96, opened 2016-02-26
Block google services users to use isolines functions

ethervoid:
Due to some TOS restrictions, google services users must not use
the HERE services which includes the isolines functions.


rafatower:
:+1:

PR #102, opened 2016-03-01
[do NOT merge] Routing function docs

ethervoid:
This are the docs for the routing functions

### DIFF
--- a/doc/routing_functions.md
+++ b/doc/routing_functions.md
@@ -1,6 +1,6 @@
 # Routing Functions
 
-The following functions provide turn-by-turn capabilities to calculate the route from one point to another using different modes, such as a car or bicycle. Currently, only the [Mapzen routing service](http://docs.cartodb.com/cartodb-platform/dataservices-api/quickstart/#using-mapzen-services) is available.
+Routing is the navigation from a defined start location to a defined end location. The calculated results are displayed as turn-by-turn directions on your map, based on the transportation mode that you specified. Routing services through CartoDB are available by requesting a single function in the Data Services API.
 
 ### cdb_route_point_to_point(_origin geometry(Point), destination geometry(Point), mode text, [options text[], units text]_)
 
--- a/doc/routing_functions.md
+++ b/doc/routing_functions.md
@@ -1,6 +1,6 @@
-# Routing functions
+# Routing Functions
 
-The following functions provide routing capabilities to calculate the route from one point to another using differents modes like car or bicycle for example. This service uses the routing service defined for the user (currently, only the Mapzen routing service is available).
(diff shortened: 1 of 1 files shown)


ethervoid:
@iriberri check the routing functions part please

iriberri:
:rainbow:

ethervoid:
Thank you very much @iriberri

ethervoid:
@csobier please review the [routing functions] (https://github.com/CartoDB/dataservices-api/commit/8d701a7b435894aae91ea71db0e8ad7e0b92e181). 

Thank you

csobier:
@ethervoid , did a check quick and applied some minor edits. I just have one question about this sentence in the introduction that might need an update:  

 "This service uses the routing service defined for the user. Currently, only the Mapzen routing service is available."

How is a service defined for the user (not by the user? As I changed it to.) It is a little bit confusing that we say it uses the routing service defined for the user, but only the Mapzen routing service is available?

ethervoid:
First of all thank you :)

You're right, right now we are only using the Mapzen routing service but this situation is going to change shortly. The sentence `defined for the user` means that the user don't have to configure the api key but this could be misleading so maybe it's better to remove it.

iriberri:
Hm, I think that the sentence could make sense. Not sure if in this scenario, but for example the Street geocoding functions works with a service or another depending on the user configuration, and this is not something that a user can change but something which is edited by sales ops or even blocked by the system itself depending on the cloud in which a user account lives.

csobier:
Okay, if you want, we could say something like "this service is defined by the user configurations that CartoDB has specified.  Currently, ..."  Otherwise, if I were a user, I would ask how I can define these services.

iriberri:
As right now there's just a routing service it doesn't make sense to struggle with this yet. :-)

ethervoid:
Sure get rid of it

csobier:
Sounds good, I removed that sentence for the time being.  Otherwise,  :+1:

ethervoid:
retest this please

csobier:
I'm not sure what happened to the edits I applied a few weeks back? Applied again and added a link to the Mapzen routing doc - in preparation for the Mapzen release- see [Docs staging PR](https://github.com/CartoDB/docs/pull/772).

doc/routing_functions.md
```diff
@@ -0,0 +1,45 @@
+# Routing Functions
+
+The following functions provide turn-by-turn capabilities to calculate the route from one point to another using different modes, such as a car or bicycle. Currently, only the [Mapzen routing service](http://docs.cartodb.com/cartodb-platform/dataservices-api/quickstart/#using-mapzen-services) is available.
```
saleiva: As mentioned, I'd remove the part talking about mapzen

PR #112, opened 2016-03-17
Routing function improvements

ethervoid:
- [x] Deleted old versioning system
- [x] 4xx responses returns empty routes despite to raise an exception
- [x] In some cases we return and empty response: one of the inputs is null, there is no generated shape for the route, etc


ethervoid:
Related #114

ethervoid:
Please @rafatower check this out :)

rafatower:
:+1:

PR #133, opened 2016-04-01
Isoline add nokia here coverage link

csobier:
@iriberri , I added a link to the HERE coverage provider - as per a Support request. I will merge unless you have any objections?


iriberri:
This can't be out because Mapzen is not released publicly. You can add it in a different PR with only the Here coverage link, but this PR is also adding Mapzen info which cannot be yet in docs.

csobier:
Yep, that's what I meant to do. Will start over. Thanks!

PR #135, opened 2016-04-04
Use retry on timeout flag for Redis connections

ethervoid:
In order to avoid some fails due some high load peaks in our Redis
databases, we are going to active this flag that retries 1 more time
in case it receives a socket.Timeout error while reading or connecting


ethervoid:
Please @iriberri check it out

iriberri:
Looks good to me :+1:

PR #140, opened 2016-04-13
Geocoder functions should be provider agnostic

ethervoid:
- [x] We have to make the geocoding functions agnostic so we need to transform the country passed when using mapzen to isocode 3
- [x] When the user pass the country and the state parameters, are concatenated to the search string.

@iriberri please take a look

### DIFF
--- a/server/extension/sql/20_geocode_street.sql
+++ b/server/extension/sql/20_geocode_street.sql
@@ -97,7 +97,10 @@ RETURNS Geometry AS $$
 
   try:
     geocoder = MapzenGeocoder(user_geocoder_config.mapzen_app_key)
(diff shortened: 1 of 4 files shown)


iriberri:
Checking

iriberri:
Looks good to me. :+1:

I thought we had the "lower regexp replace" thing abstracted in a function so that we wouldn't have it spread in every single file. I remember coding it, but I cannot find it in this repo, so it's fine.

ethervoid:
I moved the logic to the python library :)

ethervoid:
@iriberri give it another check :)

iriberri:
LGTM too, it's respecting the "city before province" hierarchy so cool!

PR #145, opened 2016-04-15
fixed admin1 name for France

nerik:
Could seem trivial but cdb_geocode_admin1_polygon actually fail to geocode french **régions**, but manages to gecode **départements** (subdivision of regions).


rafatower:
Looks ok to me.

@csobier can you please take care of this change? (not super urgent I guess)

iriberri:
Yep, you're right @nerik, these are departments:

image

Merging here. The info will be public whenever the docs site gets deployed next time. cc @csobier

csobier:
I will merge this update, and thanks for letting me know (I have a bigger WIP branch to be pushed on Monday for all the Mapzen stuff, so I'll include it there too :-)

PR #151, opened 2016-04-19
Data observatory integration

ethervoid:
- [x] Client functions for data observatory: get demographic and get segment snapshot
- [ ] Server functions for data observatory: get demographic and get segments snapshot
- [ ] All the sevice part in python library: metrics, quota, configuration, etc
- [ ] Python unit tests for this part
- [ ] Integration tests for this new functions


rafatower:
Couldn't we split this in two PR's?
- remove versioining
- everything else

ethervoid:
Could be possible yes, I'll try to do it

PR #162, opened 2016-04-22
Use redis or db data observatory quota value

ethervoid:
We don't want to update all the users in order to set a default quota so we have a default quota in our server configuration and we override it with the user-redis data when available


ethervoid:
@rafatower this one too :)

rafatower:
@jgoizueta can you please review?

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -32,15 +32,27 @@ def username(self):
     def organization(self):
         return self._orgname
 
+
 class DataObservatoryConfig(ServiceConfig):
 
+    SOFT_LIMIT_KEY = 'soft_data_observatory_limit'
+    QUOTA_KEY = 'data_observatory_quota'
```
rafatower: For keys I think it is advisable to use shorter names. There's also the thing about naming of services. How about `obs_snapshot_soft_limit`, `obs_snapshot_quota` ?

for future observatory services (making names up): `obs_augment_soft_limit`, `obs_augment_quota`, etc.

I don't think it is late to change the keys, as we can throw all the data away and start fresh. Names are quite unique and the longer version can be written in ruby/python classes handling those keys.

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -200,28 +216,6 @@ def __init__(self, redis_connection, db_conn, username, orgname=None):
         self.__parse_config(filtered_config, self._db_config)
         self.__check_config(filtered_config)
 
-    def __get_user_config(self, username, orgname):
```
jgoizueta: What happened to `GeocoderConfig`  `__get_user_config`? not needed? is it defined in base class?

ethervoid: I've refactored it, now use the self._redis_config object from the superclass in order to encapsulate all the redis_config part instead of duplicate it. This was something I missed to remove in other version

jgoizueta:
👍  looks OK to me

PR #168, opened 2016-05-03
Green/Blue deploy system for the observatory functions

### LINKED ISSUE #166: Create a mechanism to route to green-blue DO backend
Create a mechanism (another level of indirection) to route calls to the Observatory backend to the right database. It should be able:
- to store configuration of blue and green servers
- to change quickly from one another
- to override it for a set of selected users, in order to enable beta testing in production.

I'd implement it using pl/proxy (already tested, this approach will work at the expense of one more connection). I'd also implement it in the `cdb_dataservices_server` as this avoids slow config propagation.

ethervoid:
Closes #166 

Added pl/proxy loop in order to make possible the new [blue/green deployment system](http://martinfowler.com/bliki/BlueGreenDeployment.html)  for the data observatory functions


ethervoid:
@iriberri take a look please :)

iriberri:
LGTM :)

server/extension/sql/110_data_observatory.sql
```diff
@@ -1,10 +1,39 @@
...
+-- the connection string needed to connect with the current production database
+
+CREATE OR REPLACE FUNCTION cdb_dataservices_server._obs_server_conn_str(
+  username TEXT,
+  orgname TEXT)
+RETURNS text AS $$
+  plpy.execute("SELECT cdb_dataservices_server._connect_to_redis('{0}')".format(username))
+  redis_conn = GD["redis_connection_{0}".format(username)]['redis_metrics_connection']
+  plpy.execute("SELECT cdb_dataservices_server._get_obs_snapshot_config({0}, {1})".format(plpy.quote_nullable(username), plpy.quote_nullable(orgname)))
+  user_obs_snapshot_config = GD["user_obs_snapshot_config_{0}".format(username)]
```
rafatower: You're keeping the same pattern again and again and I think it is wrong. This should be:
```python
user_obs_snapshot_config = plpy.execute("SELECT cdb_dataservices_server._get_whatever(...)")
```

because of two reasons:

- you're calling a function `_get_whatever` that is not actually getting anything but causing a side effect in `GD` dictionary
- you're exposing the `GD` details to the outside world.

I hope you agree with me in that this is not a good design because of those reasons. At the time we decided not to change it to not break compatibility but we should not keep adding functions like that.

What do you think?

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -364,6 +371,12 @@ def _get_data_observatory_config(self):
         else:
             do_conf = json.loads(do_conf_json)
             self._data_observatory_monthly_quota = do_conf['monthly_quota']
+            if self._orgname and self._orgname in do_conf['connection']['whitelist']:
+                self._data_observatory_connection_str = do_conf['connection']['staging']
+            elif self._username in do_conf['connection']['whitelist']:
+                self._data_observatory_connection_str = do_conf['connection']['staging']
+            else:
+                self._data_observatory_connection_str = do_conf['connection']['production']
```
rafatower: is there a consensus about that naming? is there a defined path to promote the staging server to production?

server/extension/sql/110_data_observatory.sql
ethervoid: Ok, could be. I'm not fully agree with you but what is your proposal to change it. I don't see who we are going to encapsulate this behaviour and cache it without adding the same code in all the functions over and over again

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
ethervoid: This is going to be done by the jenkins tasks that @zenitraM is creating. That tasks also are responsible to change the config in the server which should be transparent for us.

server/extension/sql/110_data_observatory.sql
rafatower: why don't you agree?

ethervoid: I'm not agree because I don't want to copy paste the config object instantiation in all the functions, take into account that we have all the config encapsulated into Python library objects and we can't return that config object with a pl/python

rafatower: I see your point... do you see mine? I'll give it some thought

ethervoid: Yes, I'm partially agree, that's why i said that I'm not fully agree because I don't see, by now, how to extract that logic from there

rafatower:
ok, let's go ahead and I'll write my proposal in a separate PR.

zenitraM:
maybe you'll kill me by suggesting this now but what about doing it so
```
config['default_config'] = 'production'
config['production'] = ...
config['staging'] = ...
config['whitelist']['staging'] = ['foo' => 'staging', 'bar' => 'staging']
```

this way:
- `production` and `staging` are not mentioned explicitly on the code
- we can have more than one non-default environment

(In any case the current impl looks good)

ethervoid:
Right now we should go ahead with the current impl to test it and check if it works. We could add this improvements later without problem

zenitraM:
:+1:

PR #170, opened 2016-05-09
Change loopback to observatory just to call the observatory extension functions

ethervoid:
I misinterpreted the architecture/system for the blue/green deployment and the implementation is not correct, so this PR fixes the problem


ethervoid:
@rafatower review this pls :)

server/extension/cdb_dataservices_server--0.7.2--0.7.3.sql
```diff
@@ -0,0 +1,115 @@
...
+RETURNS json AS $$
+  CONNECT cdb_dataservices_server._obs_server_conn_str(username, orgname);
+  SELECT cdb_observatory.OBS_GetSegmentSnapshot(geom, geometry_level);
+$$ LANGUAGE plproxy;
+
+CREATE OR REPLACE FUNCTION cdb_dataservices_server.obs_get_segment_snapshot(
+  username TEXT,
+  orgname TEXT,
+  geom geometry(Geometry, 4326),
+  geometry_level TEXT DEFAULT '"us.census.tiger".block_group')
```
rafatower: I think these defaults should be `NULL` instead. A good moment to change them. Summoning @talos for confirmation...

server/extension/cdb_dataservices_server--0.7.2--0.7.3.sql
```diff
@@ -0,0 +1,115 @@
+--DO NOT MODIFY THIS FILE, IT IS GENERATED AUTOMATICALLY FROM SOURCES
+-- Complain if script is sourced in psql, rather than via CREATE EXTENSION
+\echo Use "ALTER EXTENSION cdb_dataservices_server UPDATE TO '0.7.3'" to load this file. \quit
+DROP FUNCTION IF EXISTS cdb_dataservices_server.obs_get_demographic_snapshot(text, text, geometry(Geometry, 4326), text, text);
```
rafatower: why dropping and not just replacing? I think the signature did not change and always that's safer.

server/extension/cdb_dataservices_server--0.7.2--0.7.3.sql
rafatower: After second thought: not in this version or it will break. Not until we go through the first blue/green deployment.

server/extension/cdb_dataservices_server--0.7.2--0.7.3.sql
ethervoid: Changed

rafatower:
:+1:

PR #172, opened 2016-05-10
On 4xx we have to return an empty response instead of an exception

### LINKED ISSUE #139: Asking for isolines from/to points far from the network leads to error.
If we try to retrieve the isocrhones to a point like this, that is a bit far from the nearest road or street

https://www.google.es/maps/place/40%C2%B022'52.6%22N+3%C2%B034'56.6%22W/@40.3816696,-3.5845506,882m/data=!3m1!1e3!4m2!3m1!1s0x0:0x0

you will get a **400 error** like this:

````
cdb_dataservices_client._cdb_isochrone(6): [cartodb_user_..._db] REMOTE ERROR: spiexceptions.InternalError: plpy.Error: There was an error trying to obtain isodistances using here maps geocoder: 400 Client Error: Bad Request for url: https://isoline.route.api.here.com/routing/7.2/calculateisoline.json?range=300&app_code=6iYGjlu41FWXYNLqjQEkjw&mode=shortest%3Bpedestrian&rangetype=time&destination=geo%2140.366407%2C-3.572695&app_id=prwe47r4wAfechukephu
````

It should give back a `null` geom in the response instead of crashing

ethervoid:
Closes #139 

We have to return an empty response instead of an exception for 4xx errors because we don't want to stop execution, and rollback everything, for this kind of errors


ethervoid:
Please @jgoizueta CR it :)

jgoizueta:
LGTM :+1:

PR #187, opened 2016-05-24
Add route_with_waypoints function

### LINKED ISSUE #106: Multiwaypoint routing
The new routing functions are awesome but it would be great to do multi waypoint routing. I know that Valhalla supports this pretty easily and returns multiple segments for the routing. I think we could join the resulting geometry for the different segments and just return that but I think being able to describe multiple waypoints would be useful. So a 

```postgresql
cdb_route_with_waypoints(waypoints geom[], method text)
```

Would be awesome to have

iriberri:
Closes #106 

Includes:
* Server
  * Python unit tests
  * Postgres tests for function signatures
* Client
  * Postgres tests for function signatures
* Global python integration tests


server/extension/sql/100_routing_helper.sql
```diff
@@ -38,12 +37,74 @@ RETURNS cdb_dataservices_server.simple_route AS $$
     orig_lat = plpy.execute("SELECT ST_Y('%s') AS lat" % origin)[0]['lat']
     orig_lon = plpy.execute("SELECT ST_X('%s') AS lon" % origin)[0]['lon']
     origin_coordinates = Coordinate(orig_lon, orig_lat)
+    waypoint_coords = [origin_coordinates]
```
ethervoid: Why have you change this function?

iriberri: Because `client.calculate_route_point_to_point` needs to receive an array of coordinates with the current code.

ethervoid: There is no problem in change the function because is in the server side and is an internal one _ so you can drop it an create another one. You have to change the mapzen routing functions but no problem because you're not going to change the signature

server/extension/cdb_dataservices_server--0.8.0--0.9.0.sql
```diff
@@ -0,0 +1,101 @@
+--DO NOT MODIFY THIS FILE, IT IS GENERATED AUTOMATICALLY FROM SOURCES
+-- Complain if script is sourced in psql, rather than via CREATE EXTENSION
+\echo Use "ALTER EXTENSION cdb_dataservices_server UPDATE TO '0.9.0'" to load this file. \quit
+DROP FUNCTION IF EXISTS cdb_dataservices_server.cdb_route_point_to_point(text, text, geometry(Geometry, 4326), geometry(Geometry, 4326), text, text[], text);
```
ethervoid: You don't need to drop the function. You're only changing the body

ethervoid: oh..you're missing the _ for the _cdb_route_point_to_point

server/extension/sql/100_routing_helper.sql
```diff
@@ -30,20 +28,18 @@ RETURNS cdb_dataservices_server.simple_route AS $$
...
-    orig_lat = plpy.execute("SELECT ST_Y('%s') AS lat" % origin)[0]['lat']
-    orig_lon = plpy.execute("SELECT ST_X('%s') AS lon" % origin)[0]['lon']
-    origin_coordinates = Coordinate(orig_lon, orig_lat)
-    dest_lat = plpy.execute("SELECT ST_Y('%s') AS lat" % destination)[0]['lat']
-    dest_lon = plpy.execute("SELECT ST_X('%s') AS lon" % destination)[0]['lon']
-    dest_coordinates = Coordinate(dest_lon, dest_lat)
-
-    resp = client.calculate_route_point_to_point(origin_coordinates, dest_coordinates, mode, options, units)
+    waypoint_coords = []
+    for idx, points in enumerate(waypoints, start=0):
```
ethervoid: idx is not needed you could use for point in waypoints

server/extension/sql/105_route_point_to_point.sql
```diff
@@ -12,7 +12,27 @@ RETURNS cdb_dataservices_server.simple_route AS $$
...
-  mapzen_plan = plpy.prepare("SELECT * FROM cdb_dataservices_server._cdb_mapzen_route_point_to_point($1, $2, $3, $4, $5, $6, $7) as route;", ["text", "text", "geometry(Point, 4326)", "geometry(Point, 4326)", "text", "text[]", "text"])
-  result = plpy.execute(mapzen_plan, [username, orgname, origin, destination, mode, options, units])
+  waypoints = [origin, destination]
+  mapzen_plan = plpy.prepare("SELECT * FROM cdb_dataservices_server._cdb_mapzen_route_with_waypoints($1, $2, $3, $4, $5, $6) as route;", ["text", "text", "geometry(Point, 4326)[]", "text", "text[]", "text"])
+  result = plpy.execute(mapzen_plan, [username, orgname, waypoints, mode, options, units])
+  return [result[0]['shape'],result[0]['length'], result[0]['duration']]
+$$ LANGUAGE plpythonu;
+
+
+CREATE OR REPLACE FUNCTION cdb_dataservices_server.cdb_route_with_waypoints(
```
ethervoid: We should change the filename to 105_route_between_points or something like that

ethervoid:
You should make that changes a test to create and drop the extension between versions and fully installation to check there is nothing left

iriberri:
Hey @ethervoid, I just pushed the changes. Thanks! I reviewed the whole upgrade/downgrade process for both server and client extensions. No issues now, and the functions that remain/are created for in each "extension update to" process is consistent with the changes we want to do.

ethervoid:
Oh good to me

PR #188, opened 2016-05-24
Add route_with_waypoints to documentation

iriberri:
Related https://github.com/CartoDB/dataservices-api/pull/187

Hi @csobier, could you review? In the PR mentioned above we're going to include a new routing function, to which you can specify a set of locations besides the origin and the destination. This is useful to force a route to go through specific locations.

cc @ethervoid


csobier:
👍 @iriberri !

iriberri:
Oops. Found a typo myself. Fixing.

iriberri:
Hi again @csobier! Sorry for the changes after your review, but I think that adding [this information](https://github.com/CartoDB/dataservices-api/commit/bf121ea3e3c7181cfdd064e9283a11aa68583f11) about how the two functions are equivalent to each other when there are just two points could be interesting. Could you please verify? :)

PR #189, opened 2016-05-25
Add routing_with_waypoints functions to Data Services API

iriberri:


iriberri:
@ethervoid could you please take a 👀?

iriberri:
After the revision, I merged the docs in the `development` so this PR now contains code + docs together.

ethervoid:
This one doesn't need review, it should just have the complete SQL files. After merge it, make a new release, one for client and other for server

server/extension/test/sql/95_route_between_points_test.sql
```diff
@@ -5,8 +5,11 @@ SELECT exists(SELECT *
               WHERE ns.nspname = 'cdb_dataservices_server'
               AND proname = 'cdb_route_point_to_point'
               AND oidvectortypes(p.proargtypes)  = 'text, text, geometry, geometry, text, text[], text');
- exists 
---------
- t
-(1 row)
 
+-- Check for routing waypoint route signatures
```
iriberri: I got a bit puzzled. The change as GitHub is showing it here looks pretty weird. I think this is a combination of the rename of files that had somehow the same contents. Despite the weirdness of this diff, `sql/95_route_between_points_test.sql` and `expected/95_route_between_points_test.out` are good.

PR #190, opened 2016-05-25
Fixed titles hierarchy

iriberri:
@csobier I noticed that due to the use of  `###` instead of  `##`  in the function names, these weren't listed in the summary sidebar. Other functions are ok, I only spotted this for routing.


csobier:
Nice! (yeah, hard to tell when you can't see what you're doing, right?).  Merge it!

PR #194, opened 2016-05-30
New release task for Makefile

ethervoid:
Added new release task in the make file to automate the new version
process:

- Move current version to old_versions folder
- Change .control file to the new version
- Create the complete SQL file for the new version
- Create empty upgrade and downgrade files

To call the new task you should pass the NEWVERSION variable. Eg:

make release NEWVERSION=0.x.0

### DIFF
--- a/server/extension/Makefile
+++ b/server/extension/Makefile
@@ -2,39 +2,47 @@
 # Once a version is released, it is not meant to be changed. E.g: once version 0.0.1 is out, it SHALL NOT be changed.
 EXTENSION = cdb_dataservices_server
 EXTVERSION = $(shell grep default_version $(EXTENSION).control | sed -e "s/default_version[[:space:]]*=[[:space:]]*'\([^']*\)'/\1/")
-
 # The new version to be generated from templates
+SED = sed
+ERB = erb
+REPLACEMENTS = -i 's/$(EXTVERSION)/$(NEW_VERSION)/g'
 NEW_EXTENSION_ARTIFACT = $(EXTENSION)--$(EXTVERSION).sql
-OLD_VERSIONS = $(wildcard old_versions/*.sql)
+
+REGRESS = $(notdir $(basename $(wildcard test/sql/*test.sql)))
+TEST_DIR = test/
+REGRESS_OPTS = --inputdir='$(TEST_DIR)' --outputdir='$(TEST_DIR)'
 
 # DATA is a special variable used by postgres build infrastructure
 # These are the files to be installed in the server shared dir,
 # for installation from scratch, upgrades and downgrades.
 # @see http://www.postgresql.org/docs/current/static/extend-pgxs.html
-DATA =  $(NEW_EXTENSION_ARTIFACT) \
-  $(OLD_VERSIONS) \
-  cdb_dataservices_server--0.9.0--0.10.0.sql \
-  cdb_dataservices_server--0.10.0--0.9.0.sql
-
-REGRESS = $(notdir $(basename $(wildcard test/sql/*test.sql)))
-TEST_DIR = test/
-REGRESS_OPTS = --inputdir='$(TEST_DIR)' --outputdir='$(TEST_DIR)'
+DATA = $(EXTENSION)--*.sql
+OLD_VERSIONS = $(wildcard old_versions/*.sql)
+SOURCES_DATA_DIR = sql/
+SOURCES_DATA = $(wildcard sql/*.sql)
 
 # postgres build stuff
 PG_CONFIG = pg_config
 PGXS := $(shell $(PG_CONFIG) --pgxs)
 include $(PGXS)
 
-SOURCES_DATA_DIR = sql/
-
-SOURCES_DATA = $(wildcard sql/*.sql)
-
 $(NEW_EXTENSION_ARTIFACT): $(SOURCES_DATA)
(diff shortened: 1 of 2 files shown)


server/extension/Makefile
```diff
@@ -3,6 +3,15 @@
 EXTENSION = cdb_dataservices_server
 EXTVERSION = $(shell grep default_version $(EXTENSION).control | sed -e "s/default_version[[:space:]]*=[[:space:]]*'\([^']*\)'/\1/")
 
+SED = sed
+REPLACEMENTS = -i 's/$(EXTVERSION)/$(NEWVERSION)/g'
```
rafatower: I'd change NEWVERSION to NEW_VERSION for readability's sake

server/extension/Makefile
```diff
@@ -3,6 +3,15 @@
...
+define UPGRADE_DOWNGRADE_FILES_HEADER
+--DO NOT MODIFY THIS FILE, IT IS GENERATED AUTOMATICALLY FROM SOURCES
+-- Complain if script is sourced in psql, rather than via CREATE EXTENSION
+\echo Use "ALTER EXTENSION cdb_dataservices_server UPDATE TO ''" to load this file. \quit
+endef
+export UPGRADE_DOWNGRADE_FILES_HEADER
```
rafatower: why not in a file? it is really ugly in an envar

server/extension/Makefile
```diff
@@ -11,10 +20,7 @@ OLD_VERSIONS = $(wildcard old_versions/*.sql)
 # These are the files to be installed in the server shared dir,
 # for installation from scratch, upgrades and downgrades.
 # @see http://www.postgresql.org/docs/current/static/extend-pgxs.html
-DATA =  $(NEW_EXTENSION_ARTIFACT) \
-  $(OLD_VERSIONS) \
-  cdb_dataservices_server--0.8.0--0.9.0.sql \
-  cdb_dataservices_server--0.9.0--0.8.0.sql
+DATA = $(EXTENSION)--*.sql \
```
rafatower: So you're grabbing all the files `cdb_dataservices_server--*.sql` in the current dir. Is that what you want?

server/extension/Makefile
```diff
@@ -35,6 +41,18 @@ $(NEW_EXTENSION_ARTIFACT): $(SOURCES_DATA)
...
 
+release: $(EXTENSION).control $(SOURCES_DATA)
+	test -n "$(NEWVERSION)"  # $$NEWVERSION VARIABLE MISSING. Eg. make release NEWVERSION=0.x.0
+	mv *.sql old_versions
+	$(SED) $(REPLACEMENTS) $(EXTENSION).control
+	cat $(SOURCES_DATA_DIR)/*.sql > $(EXTENSION)--$(NEWVERSION).sql
+	touch $(EXTENSION)--$(EXTVERSION)--$(NEWVERSION).sql
+	touch $(EXTENSION)--$(NEWVERSION)--$(EXTVERSION).sql
+	echo "$$UPGRADE_DOWNGRADE_FILES_HEADER" > $(EXTENSION)--$(EXTVERSION)--$(NEWVERSION).sql
+	echo "$$UPGRADE_DOWNGRADE_FILES_HEADER" > $(EXTENSION)--$(NEWVERSION)--$(EXTVERSION).sql
```
rafatower: I don't know what you want here but doesn't seem right.

server/extension/Makefile
ethervoid: I think about it but yes I'll change it

server/extension/Makefile
```diff
@@ -35,6 +41,18 @@ $(NEW_EXTENSION_ARTIFACT): $(SOURCES_DATA)
...
+	$(SED) -i 's/UPDATE TO \x27\x27/UPDATE TO \x27$(NEWVERSION)\x27/g' $(EXTENSION)--$(EXTVERSION)--$(NEWVERSION).sql
+	$(SED) -i 's/UPDATE TO \x27\x27/UPDATE TO \x27$(EXTVERSION)\x27/g' $(EXTENSION)--$(NEWVERSION)--$(EXTVERSION).sql
+
```
rafatower: same here

server/extension/Makefile
ethervoid: I thought too i decided to add inside, despite is ugly as you say, because I don't want an empty and useless file in the directory or make a templates directory just for the headers of a file

server/extension/Makefile
ethervoid: Yes, in the current directory should be only 3 sql files, the whole file and the upgrade/downgrade ones

server/extension/Makefile
ethervoid: I'm adding the headers to the new files without version, then I will change the empty part for the current version. I don't know if there are templates in the makefiles

jgoizueta: but isn't `$(EXTENSION)--$(NEWVERSION).sql` the one file "not to be modified"? these two files must be populated manually, aren't they?

ethervoid: Yes but we don't want the users to modify them. We show them as automatically generated files

rafatower:
Ok

Just one minor thing: `>>` is used to append and I'm not sure why would you append rather than overwriting that

ethervoid:
True, fixed

PR #195, opened 2016-05-30
OBS_GetMeasureById

ethervoid:
- New function obs_getmeasurebyid(text, text, text, text) from version 0.0.5 of observatory extension


ethervoid:
Please @rafatower CR this :)

rafatower:
:+1:

PR #202, opened 2016-06-13
201 read quota redis both formats

rafatower:
@iriberri and @jgoizueta please review

my python skills were a bit rusted :laughing:


iriberri:
Aren't you updating the version of the Python package?

Besides that, it looks good to me. And thanks for the docs! :)

rafatower:
Mmm, yep, you're right, I have to update the version of the package. Thanks for pointing out.

PR #205, opened 2016-06-14
204 write zero padded keys

rafatower:
@jgoizueta please review

### DIFF
--- a/server/lib/python/cartodb_services/cartodb_services/metrics/user.py
+++ b/server/lib/python/cartodb_services/cartodb_services/metrics/user.py
@@ -105,7 +105,9 @@ class UserMetricsService:
     def __increment_organization_uses(self, service_type, metric, date, amount):
         redis_prefix = self.__parse_redis_prefix("org", self._orgname,
                                                  service_type, metric, date)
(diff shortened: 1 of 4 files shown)


server/lib/python/cartodb_services/cartodb_services/metrics/user.py
```diff
@@ -8,6 +8,7 @@ class UserMetricsService:
     SERVICE_GEOCODER_NOKIA = 'geocoder_here'
     SERVICE_GEOCODER_CACHE = 'geocoder_cache'
     SERVICE_HERE_ISOLINES = 'here_isolines'
+    DAY_OF_MONTH_ZERO_PADDED = '%d'
```
jgoizueta: Should use the constant here too: https://github.com/CartoDB/dataservices-api/blob/204-write-zero-padded-keys/server/lib/python/cartodb_services/cartodb_services/metrics/user.py#L89

jgoizueta:
:+1:

server/lib/python/cartodb_services/cartodb_services/metrics/user.py
rafatower: right!

PR #208, opened 2016-06-21
207 fix test if obs search is ok

rafatower:
Make the test_if_obs_search_is_ok a bit more robust by narrowing the
query to exactly what is expected, as suggested by talos.

@iriberri or @jgoizueta  can you please review? (mostly informational)

Notes:
1. I had to add the url quoting to make it work with a percentage sign (`%`) in the sql query.
2. I don't like this patch, the same way I do not like the fact that it's a bit blurred in the test code who's responsible for quoting url params or composing them (I tried to get it fixed in the test helper, but it is a bit messed up).
3. I also dislike the test runner, which makes it pretty hard to plug a debugger in.

So to sum up: we need a quick fix for the integration tests to keep running but also consider improving that code.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -2,9 +2,11 @@
 This are the automatic integration tests for geocoder api (both client and server)
 
 ### Usage
-In order to execute the tests you have to execute the `run_tests` python script:
+In order to execute the tests you have to execute the `run_tests.py` python script:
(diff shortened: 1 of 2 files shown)


iriberri:
:+1:

PR #210, opened 2016-06-29
Add OBS_AugmentTable and OBS_GetTable

iriberri:
Some notes/details:

##### Notes regarding style
* I renamed a lot of variables in the client side since last review, I hope this is more clear now
* Python `.format` lines are extra long because of the dictionary I'm using for variable substitution
* There are several lines to build strings that we need to use them in SQL. Mostly, we need a string with the name of the columns separated by commas, and we need a string with a list of names and types (to retrieve the RECORDs from the remote end)

##### Notes regarding functionality
* I'm creating temporary tables (for augment function) with a specific prefix and an epoch timestamp. These are **unlogged** and dropped afterwards. Creating temp pg tables would have the inconvenience that we need to know which temporal schema name is the session using. Apparently this is configurable but I never found it.
* In AugmentTable, the function will fail if there's already a column with the name of the measure(s) requested (except for functions that operate over  `the_geom`). We could workaround this by setting different column names (with a timestamp, sufix or whatever) but I don't think this improves anything
* In GetTable, the new table will have **all** columns from the user table except in the case where `the_geom` gets updated or there are columns in the user table with the same name. I will only select columns from the user table which do not collide with the Observatory results (left join). For this I'm making some queries to `pg_attribute`

##### Usage examples
```sql
select OBS_AugmentTable('<TABLE_NAME_HERE>', 'GetMeasure', '{"tag_name":["total_pop"]}'::json);
````

```sql
select _OBS_GetTable('<TABLE_NAME_HERE>','<OUTPUT_TABLE_NAME_HERE>', 'GetMeasure', '{"tag_name":["total_pop", "pop_16_over"]}'::json);
````

Note: Current the functions will only work with 'GetMeasure' and one or more tags: `total_pop` and/or `pop_16_over`.

### CODE REVIEW

**Comment by iriberri** [comment]
Please @rafatower, could you take a look? :-)

**Comment by rafatower** [comment]
test broken :S

**Comment by rafatower** [comment]
It's a change in platform setup:

```
14:48:03 + source /home/ubuntu/.virtualenvs/dataservices-api/bin/activate
14:48:03 /tmp/hudson4016317693051871836.sh: line 4: /home/ubuntu/.virtualenvs/dataservices-api/bin/activate: No such file or directory
14:48:03 Build step 'Execute shell' marked build as failure
```

File: client/sql/20_table_augmentation.sql [hunk 1]
```diff
@@ -0,0 +1,231 @@
+CREATE TYPE cdb_dataservices_client.ds_fdw_metadata as (schemaname text, tabname text, servername text);
+CREATE TYPE cdb_dataservices_client.ds_return_metadata as (colnames text[], coltypes text[]);
+
+CREATE OR REPLACE FUNCTION cdb_dataservices_client.OBS_GetTable(table_name text, output_table_name text, function_name text, params json)
```
rafatower: I don't think this function shall be "public" yet, better to have it with the underscore prepended

client/sql/20_table_augmentation.sql
```diff
@@ -0,0 +1,231 @@
...
+  SELECT current_database() INTO dbname;
+
+  SELECT cdb_dataservices_client._OBS_GetTable(username, orgname, user_db_role, user_schema, dbname, table_name, output_table_name, function_name, params) INTO result;
+
+  RETURN result;
+END;
+$$ LANGUAGE 'plpgsql' SECURITY DEFINER;
+
+
+CREATE OR REPLACE FUNCTION cdb_dataservices_client.OBS_AugmentTable(table_name text, function_name text, params json)
```
rafatower: I don't think this function shall be "public" yet, better to have it with the underscore prepended

client/sql/20_table_augmentation.sql
```diff
@@ -0,0 +1,231 @@
...
+        colnames_arr = ds_return_metadata[0]["colnames"]
+        coltypes_arr = ds_return_metadata[0]["coltypes"]
+
+        # Prepare column and type strings required in the SQL queries
+        colnames = ','.join(colnames_arr)
+        columns_with_types_arr = [colnames_arr[i] + ' ' + coltypes_arr[i] for i in range(0,len(colnames_arr))]
+        columns_with_types = ','.join(columns_with_types_arr)
+
+
+        # Obtain metadata for FDW connection
```
rafatower: The purpose here is to instruct obs to establish a FDW. How about...
```python
# Instruct the OBS server side to establish a FDW
# the metadata is obtained as well in order to be able to tell OBS to free resources when done.
```
?

client/sql/20_table_augmentation.sql
```diff
@@ -0,0 +1,231 @@
...
+        wiped = plpy.execute("SELECT cdb_dataservices_client._OBS_DisconnectUserTable({username}::text, {orgname}::text, '{server_schema}'::text, '{server_table_name}'::text, '{fdw_server}'::text)"
+            .format(username=plpy.quote_nullable(username), orgname=plpy.quote_nullable(orgname), server_schema=server_schema, server_table_name=server_table_name, fdw_server=server_name))
+
+        # Prepare table to receive augmented results
+        for idx, column in enumerate(colnames_arr):
+            if colnames_arr[idx] is not 'the_geom':
+                plpy.execute('ALTER TABLE "{user_schema}".{table_name} ADD COLUMN {column_name} {column_type}'
+                    .format(user_schema=user_schema, table_name=table_name, column_name=colnames_arr[idx], column_type=coltypes_arr[idx]))
+
+        # Populate a new table with the augmented results
```
rafatower: in this case it is not a new table

client/sql/20_table_augmentation.sql
```diff
@@ -0,0 +1,231 @@
...
+        return True
+    except Exception as e:
+        plpy.warning('Error trying to augment table {0}'.format(e))
+        # Wipe user FDW data from the server in case of failure if the table was connected
+        if server_table_name:
+            # Wipe local temporary table
+            plpy.execute('DROP TABLE IF EXISTS "{user_schema}".{temporary_table_name}'
+                .format(user_schema=user_schema, table_name=table_name, temporary_table_name=temporary_table_name))
+
+            wiped = plpy.execute("SELECT cdb_dataservices_client._OBS_DisconnectUserTable({username}::text, {orgname}::text, '{server_schema}'::text, '{server_table_name}'::text, '{fdw_server}'::text)"
```
rafatower: isn't this already done above?

iriberri: Yep, it is dropping the table in the regular flow, but in case of any exception I'm triggering the function again to make sure the resources are free. I was using a `finally` clause before (we already discussed about this) but due to it I had some issues with the RETURNs, as whatever is the outcome it will always pass through there.

client/sql/20_table_augmentation.sql
```diff
@@ -0,0 +1,231 @@
...
+            .format(username=plpy.quote_nullable(username), orgname=plpy.quote_nullable(orgname), user_db_role=user_db_role, user_schema=user_schema, dbname=dbname, table_name=table_name))
+
+        server_schema = ds_fdw_metadata[0]["schemaname"]
+        server_table_name = ds_fdw_metadata[0]["tabname"]
+        server_name = ds_fdw_metadata[0]["servername"]
+
+        # Create temporary table with the augmented results
+        plpy.execute('CREATE UNLOGGED TABLE "{user_schema}".{temporary_table_name} AS '
+            '(SELECT {columns}, cartodb_id '
+            'FROM cdb_dataservices_client._OBS_GetProcessedData({username}::text, {orgname}::text, \'{schema}\'::text, \'{table_name}\'::text, \'{function_name}\'::text, \'{params}\'::json) as results({columns_with_types}, cartodb_id int) )'
```
rafatower: `_OBS_GetProcessedData` doesn't strike me as a good name yet.

How about `_OBS_FetchJoinFdwTableData`? Actually it is processing on the fly, then getting the dynamic records through pl/proxy. I can't come up with a better name.

rafatower: I'd split this line into several parts

rafatower: I'd avoid quoting manually: ` \'{schema}\'::text`. Better using `plpy.quote_*` methods

client/sql/20_table_augmentation.sql
```diff
@@ -0,0 +1,231 @@
...
+            .format(columns=colnames, username=plpy.quote_nullable(username), orgname=plpy.quote_nullable(orgname), user_schema=user_schema, schema=server_schema, table_name=server_table_name, function_name=function_name, params=params, columns_with_types=columns_with_types, temporary_table_name=temporary_table_name))
```
rafatower: I'd split this into several lines

server/extension/sql/125_data_observatory_table_augment.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TYPE cdb_dataservices_server.ds_fdw_metadata as (schemaname text, tabname text, servername text);
+
+CREATE TYPE cdb_dataservices_server.ds_return_metadata as (colnames text[], coltypes text[]);
+
+CREATE OR REPLACE FUNCTION cdb_dataservices_server.OBS_ConnectUserTable(username text, orgname text, user_db_role text, input_schema text, dbname text, table_name text)
+RETURNS cdb_dataservices_server.ds_fdw_metadata AS $$
+    return plpy.execute("SELECT * FROM cdb_dataservices_server._OBS_ConnectUserTable({0}::text, {1}::text, '{2}'::text, '{3}'::text, '{4}'::text, '{5}'::text)"
```
rafatower: I'd apply the same thing regarding named placeholders here, rather than `{1}, {2}, {3}, etc.`

rafatower: I guess this level of indirection is in preparation for quotas and the like, right?

rafatower:
I left a bunch of comments but it looks pretty good

iriberri:
Thanks @rafatower, gonna apply them 'cause them all make sense :-)

client/sql/20_table_augmentation.sql
rafatower: this is something to improve. I think the issue with the returns and the finally and the errors is that plpy.error does not return from a function which continues executing stuff until it finishes.

but when you find something weird it is better to understand what is happening under the hoods, normally isolating the problem in a small toy sample. Are you able to write such example? I remember the discussion but haven't ever seen the "faulty" code.

since it's an improvement, I'm not pushing for it right now but I'd be happy to have a ticket in the backlog for it.

server/extension/sql/125_data_observatory_table_augment.sql
iriberri: Yes, this function is the counterpart of the plpython functions we have in the server which typically manage the quota. Here's not needed yet but I'm still keeping the structure. Otherwise just a simple plproxy would work.

client/sql/20_table_augmentation.sql
iriberri: :+1: Yes, I can reproduce it easily.

The main idea is that I was adding the return inside the finally so... it would always return that no matter what happened. I guess I can play with status variables or just recheck again because I might have missed something when handling the exceptions. From the docs:

> A finally clause is always executed before leaving the try statement, whether an exception has occurred or not. When an exception has occurred in the try clause and has not been handled by an except clause (or it has occurred in a except or else clause), it is re-raised after the finally clause has been executed. The finally clause is also executed “on the way out” when any other clause of the try statement is left via a break, continue or return statement.

iriberri:
Retest this, please

iriberri:
Retest this, please

iriberri:
Retest this, please

iriberri:
I just removed the observatory dependency.

PR #213, opened 2016-07-05
209 Exposes geocoder providers in public geocoder functions

### LINKED ISSUE #209: Expose particular geocoder functions in the public API
In particular the following functions:
```
cdb_dataservices_server._cdb_here_geocode_street_point
cdb_dataservices_server._cdb_google_geocode_street_point
cdb_dataservices_server._cdb_mapzen_geocode_street_point
```

Quota/config checks should still apply but let for instance a user with `here` by default use explicitly `mapzen` or `here`.

See https://github.com/CartoDB/dataservices-api/blob/development/server/extension/sql/20_geocode_street.sql

**UPDATE**: `cdb_dataservices_server._cdb_mapzen_geocode_street_point` is the priority

iriberri:
Closes #209

Some details about the solution:

**Client (pg)**
* Added three new functions in the client for the three different providers by using the templates system.

**Server (pg)**
* Added three new public functions for each provider:
  * `cdb_here_geocode_street_point`: Checks if user has HERE configured as default geocoder. Otherwise it raises.
  *  `cdb_google_geocode_street_point`: Checks if user has Google configured as default geocoder. Otherwise it raises.
  * `cdb_mapzen_geocode_street_point`: Does not perform any check in advance. It will use a specific function to get the Mapzen configuration from Python.
 
* Edited `_cdb_mapzen_geocode_street_point` for it to use a new function to obtain Mapzen configuration. Otherwise users would have heremaps/google enabled and would not be able to go through the Mapzen functions.

* Added `_get_mapzen_geocoder_config` which uses `MapzenGeocoderConfig` from Python.

* The new server functions can be found explicitly in [the upgrade file](https://github.com/CartoDB/dataservices-api/pull/213/files#diff-23a0f49f581af8db827e18d7a73b2630).

**Server (python)**
* Added new class `MapzenGeocoderConfig` in config. It takes the mapzen geocoder API and (general) quota limit set up in CDB_Conf.


server/extension/sql/15_config_helper.sql
```diff
@@ -12,6 +12,20 @@ RETURNS boolean AS $$
...
 
+CREATE OR REPLACE FUNCTION cdb_dataservices_server._get_mapzen_geocoder_config(username text, orgname text)
+RETURNS boolean AS $$
+  cache_key = "user_mapzen_geocoder_config_{0}".format(username)
+  if cache_key in GD:
+    return False
+  else:
+    from cartodb_services.metrics import MapzenGeocoderConfig
+    plpy.execute("SELECT cdb_dataservices_server._connect_to_redis('{0}')".format(username))
+    redis_conn = GD["redis_connection_{0}".format(username)]['redis_metadata_connection']
```
rafatower: do we need redis for this?

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -128,6 +128,60 @@ def monthly_quota(self):
     def period_end_date(self):
         return self._period_end_date
 
+# Explicit class for the geocoder configuration for Mapzen
+# which does not require users to be configured to use the service
+class MapzenGeocoderConfig(ServiceConfig):
+
+    PERIOD_END_DATE = 'period_end_date'
+
+    def __init__(self, redis_connection, db_conn, username, orgname=None):
```
rafatower: the `redis_connection` is only needed because it inherits from `ServiceConfig`. Are we using anything from the superclass that justifies creating/passing a redis connection?

BTW, not your fault cause you're following an existing design, but worth giving some thought to it: https://www.thoughtworks.com/es/insights/blog/composition-vs-inheritance-how-choose

server/extension/sql/15_config_helper.sql
iriberri: I suppose we want to keep track of usage, so in this case we need it.

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
rafatower: okay, redis is used for metrics so it is OK.

server/extension/sql/15_config_helper.sql
rafatower: ok

iriberri:
Some comments on my tests in staging:

* For a user with heremaps configured by default:
  * here function works
  * mapzen function works
  * generic function works
  * google function raises

Also, if I downgrade the server to 0.10.0 the old generic function keeps working (but the new functions raise a "function does not exist" error).


It's looking good! \o/

iriberri:
The metrics stored in Redis are also working as I would expect:

````
127.0.0.1:6379[5]> zscore user:ethervoid-ded04:geocoder_mapzen:total_requests:201607 05
"1"
```

iriberri:
Integration tests are working for the new mapzen function but an observatory function is raising an error:

````
10:50:25 FAIL: test_if_obs_get_available_boundaries_is_ok (test.integration.test_data_observatory_functions.TestDataObservatoryFunctions)
10:50:25 ----------------------------------------------------------------------
10:50:25 Traceback (most recent call last):
10:50:25   File "/home/ubuntu/workspace/CartoDB-DATASERVICES-API-Integration-testing/test/integration/test_data_observatory_functions.py", line 150, in test_if_obs_get_available_boundaries_is_ok
10:50:25     assert_equal(result['boundary_id'], 'whosonfirst.wof_continent_geom')
10:50:25 AssertionError: u'us.census.tiger.block_group_clipped' != 'whosonfirst.wof_continent_geom'
```

rafatower:
:+1: the obs function works in produciton against the last released stuff.

PR #215, opened 2016-07-06
Fix bug on exception raise for Mapzen geocoder config

iriberri:
When I was testing the new code I didn't experience any issue with configs as I had it set up correctly, but today I noticed about this. `key` is not defined in that scope at all so I'm editing the tuple a bit.

Please @rafatower take a look. This is what I was mentioning before.


iriberri:
Merging this. Really silly change!

PR #220, opened 2016-07-11
Bump client to 0.10.0 and server to 0.13.0

iriberri:


iriberri:
@ethervoid could you help me take a look here? I didn't commit directly to develop so that we could take a look at this after all the versioning mess.

ethervoid:
Looks good to me

iriberri:
I have actually found some issues with some tests. I should remove the functions before the types, otherwise an error will be raised due to the functions needed the type (return types). The other approach is to use a `drop cascade` but sounds dangerous. Uploading a new commit right now!

ethervoid:
Yup, I've missed that

iriberri:
I'm gonna include the details for the dump version in observatory too that I have prepared. If we don't go ahead with that we can just revert the commit.

PR #225, opened 2016-07-15
Fix client address recognition in augment functions

iriberri:
Moving the detection of the user IP a step before. Previously the function was run in the inner observatory functions, and therefore the IP detected was the DS server one, not the one from the client.


iriberri:
The function in observatory is also updated to receive the addr_host as a parameter, instead of obtaining it internally (which was returning a nice 127.0.0.1 ;-)): https://github.com/CartoDB/observatory-extension/pull/127/files#diff-deb0e280f1fa17a0da59c028be014712R1

iriberri:
@ethervoid please, could you CR? It's working, but for you to :eyes:

PR #230, opened 2016-07-19
Add index to temporary table in AugmentTable

iriberri:


iriberri:
Hey @ethervoid could you review? As always, it's been already tested but it's time to move it to production :-) I think I'm finally just going to release this simple fix on its own.

ethervoid:
Looks good to me 👍

PR #232, opened 2016-07-22
Fix isolines range bug

ethervoid:
Fixes Cartodb/cartodb#9076


server/lib/python/cartodb_services/setup.py
```diff
@@ -10,7 +10,7 @@
 setup(
     name='cartodb_services',
 
-    version='0.7.0',
+    version='0.7.0.1',
```
iriberri: Ooops :)

ethervoid:
@iriberri CR please

PR #237, opened 2016-07-25
Replace all 'area' defaults for "normalize" with NULL. 

### LINKED ISSUE #235: Normalize for getmeasure should default to NULL
We handle the default for `normalize` further down the line by assuming it's `NULL`, but it's being defaulted to `area` in PL/Proxy:

https://github.com/CartoDB/dataservices-api/blob/d57856367a0d2a131d6e0285d7e5fec673a39f66/server/extension/sql/110_data_observatory_augmentation.sql#L226

This produces unexpected behavior (normalization by area) which is different than expected according to documentation for polygons: https://carto.com/docs/carto-engine/data/measures-functions/#obsgetmeasurepolygon-geometry-measureid-text

> for measures that are are sums (e.g. population) the default normalization is ‘none’ and response comes back as a raw value. Other options are ‘denominator’, which will use the denominator specified in the Data Catalog (optional)

A simple one-line change should suffice, but it might cause test failures down the line.  CC @iriberri , let me know if you want me to make a PR right now with that one line change.

talos:
Fixes #235.

CR @iriberri ?  I think I caught everything, but might have missed some.

After this is done, the analysis pipeline should have the desired default behavior when obtaining the measure for a polygon (which is to find the actual number for that polygon instead of a rate).

### DIFF
--- a/client/renderer/interface.yaml
+++ b/client/renderer/interface.yaml
@@ -147,13 +147,13 @@
   params:
     - { name: geom,  type: "geometry(Geometry, 4326)" }
     - { name: time_span,  type: "text", default: "'2009 - 2013'::text" }
-    - { name: geometry_level, type: text, default: "'\"us.census.tiger\".block_group'::text" }
+    - { name: geometry_level, type: text, default: 'NULL' }
 
 - name: obs_get_segment_snapshot
   return_type: json
   params:
     - { name: geom,  type: "geometry(Geometry, 4326)" }
-    - { name: geometry_level, type: text, default: "'\"us.census.tiger\".census_tract'::text" }
+    - { name: geometry_level, type: text, default: 'NULL' }
 
 - name: obs_getdemographicsnapshot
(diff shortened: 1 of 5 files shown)


iriberri:
Hi @talos! Thanks :)

I'm going ahead and adding/reviewing some more commits. We also had default values for "intersects" and the geometry ID, and I'm totally convinced of the fact that these defaults should not live here because DS API should be totally agnostic and should not include this "extra logic". I'll be still committing changes after I write this comment, could you please take a look when you can and confirm that the other defaults that I'm getting rid of are covered in the Observatory extension? (Like I won't break anything in the case OBS is assuming that DS API is filling in the gaps).

talos:
@ethervoid Any update on this?  This is causing some confusion for users.

ethervoid:
I'll work on it as soon as possible.

talos:
👍

PR #244, opened 2016-07-29
Remove 1 row limit for isolines functions

### LINKED ISSUE #245: Isolines limited to 1 rows
We have the isolines functions limited to return only 1 row so when a user try to get multiple range isolines, only get the first row

ethervoid:
Closes #245


iriberri:
LGTM:+1:

PR #248, opened 2016-08-01
Improve logger capability

ethervoid:
- [x] New logger with: plpy, rollbar integration and file
- [x] Added min log level to notify as config option
- [x] Server config to define dataservices environment for: log, third party servers, etc
- [x] Added logger to the SQL functions
- [x] Raise exception on events that should not be logged like reach the quota limit


ethervoid:
@rafatower take a first look. I've removed the upgrade/downgrade files so should be easier to make a CR

server/extension/sql/100_routing_helper.sql
```diff
@@ -17,19 +17,22 @@ RETURNS cdb_dataservices_server.simple_route AS $$
   from cartodb_services.mapzen.types import polyline_to_linestring
   from cartodb_services.metrics import QuotaService
   from cartodb_services.tools import Coordinate
+  from cartodb_services.tools import Logger
 
   redis_conn = GD["redis_connection_{0}".format(username)]['redis_metrics_connection']
   user_routing_config = GD["user_routing_config_{0}".format(username)]
 
+  logger = Logger(user_routing_config)
```
rafatower: why does the logger need the user routing config to be instantiated?

ethervoid: Because the logger configuration is inside the DB and the config object has all what is needed. I'd love not to do this way but the other way is pass the DB object

server/extension/sql/30_admin0.sql
```diff
@@ -20,12 +22,10 @@ RETURNS Geometry AS $$
...
     except BaseException as e:
-      import sys, traceback
-      type_, value_, traceback_ = sys.exc_info()
+      import sys
       quota_service.increment_failed_service_use()
-      error_msg = 'There was an error trying to geocode using admin0 geocoder: {0}'.format(e)
-      plpy.notice(traceback.format_tb(traceback_))
-      plpy.error(error_msg)
+      logger.error('Error trying to geocode admin0 polygon', sys.exc_info())
+      raise e
```
rafatower: I'm concerned about confusing users with error messages. At least they were used to see 'There was an error trying to geocode blah blah'

ethervoid: I was thinking to change the raise e for raise Exception('There was an error trying to...') and avoid more info to avoid leak of sensitive information.

server/lib/python/cartodb_services/cartodb_services/mapzen/isolines.py
```diff
@@ -85,11 +86,7 @@ def calculate_isodistance(self, origin, transport_mode, distance_range):
     def calculate_isoline(self, origin, costing_model, isorange, upper_rmax, cost_variable, unit_factor=1.0):
 
         # NOTE: not for production
-        #logging.basicConfig(level=logging.DEBUG, filename='/tmp/isolines.log')
-        #logging.basicConfig(level=logging.DEBUG)
-        logging.debug('origin = %s' % origin)
-        logging.debug('costing_model = %s' % costing_model)
-        logging.debug('isorange = %d' % isorange)
+        self._logger.debug('Calculate isoline', data={"origin": origin, "costing_model": costing_model, "isorange": isorange})
```
rafatower: what default log level do you plan for production?

ethervoid: warning https://github.com/CartoDB/dataservices-api/pull/248/files/ed5cf25e9cb03c9c7d227c4656ef9f5a4c800db7#diff-1c9765409ffd2cf95aa36ab177efe7b7R491

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -33,6 +37,23 @@ def username(self):
     def organization(self):
         return self._orgname
 
+    @property
+    def rollbar_api_key(self):
```
rafatower: this doesn't fit in metrics very well

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -162,6 +186,7 @@ def __parse_config(self, filtered_config, db_config):
         if not self._isolines_provider:
             self._isolines_provider = self.DEFAULT_PROVIDER
         self._geocoder_provider = filtered_config[self.GEOCODER_PROVIDER_KEY].lower()
+        self._period_end_date = date_parse(filtered_config[self.PERIOD_END_DATE])
```
rafatower: why was this changed?

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
ethervoid: I don't think this config is in metrics anymore. Needs refactor but by now I want to add feature not mix refactor and new feature. The two hats syndrome xD

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
ethervoid: I put above. It was below the if, just to clarify. Is not a change

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -400,11 +424,23 @@ def __init__(self, db_conn, username, orgname):
...
+        self._get_server_config()
         self._get_here_config()
         self._get_mapzen_config()
         self._get_logger_config()
         self._get_data_observatory_config()
 
+    def _get_server_config(self):
+        server_config_json = self._get_conf('server_conf')
+        if not server_config_json:
+            self._server_environment = 'production'
```
rafatower: why is `'production'` the default?

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -451,6 +487,15 @@ def _get_logger_config(self):
...
             self._geocoder_log_path = logger_conf['geocoder_log_path']
+            self._rollbar_api_key = None
+            self._min_log_level = 'warning'
+            self._log_file_path = None
+            if 'min_log_level' in logger_conf:
+                self._min_log_level = logger_conf['min_log_level']
+            if 'rollbar_api_key' in logger_conf:
+                self._rollbar_api_key = logger_conf['rollbar_api_key']
+            if 'log_file_path' in logger_conf:
+                self._log_file_path = logger_conf['log_file_path']
```
rafatower: I don't see why this is coupled with here or any other provider

server/lib/python/cartodb_services/cartodb_services/metrics/log.py
```diff
@@ -4,17 +4,17 @@
 import re
 
 
-class LoggerFactory:
+class MetricsLoggerFactory:
```
rafatower: Here I see I'm missing changes, probably from an earlier PR already merged

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
ethervoid: Because is the default environment right now if the property is missing.

server/lib/python/cartodb_services/cartodb_services/metrics/log.py
rafatower: was this file used before? Seems like it was first committed on March

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
ethervoid: This part is just for logger (metrics and log) is not coupled. The better scenario is have it out of here but I have to create a new function in SQL part to create the config and pass the db connection and so on

server/lib/python/cartodb_services/cartodb_services/metrics/log.py
ethervoid: This is the metrics logger used by systems to gather metrics from the services

rafatower: Okay, I see we're talking here about different loggers. The resulting files were supposed to be consumed by ES.

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
ethervoid: It's possible to pass the plpy object and create the logger from there...

server/lib/python/cartodb_services/cartodb_services/tools/log.py
```diff
@@ -0,0 +1,108 @@
...
+        if data:
+            extra_data['data'] = data
+        else:
+            extra_data['data'] = ''
+
+        return extra_data
+
+    def _setup_log_file_config(self, log_file_path):
+        format_str = "%(asctime)s %(name)-12s %(levelname)-8s %(message)s %(data)s %(exception_type)s %(exception_message)s %(exception_traceback)s"
+        logging.basicConfig(filename=log_file_path, format=format_str, level=self._config.min_log_level.upper())
```
rafatower: I wouldn't use that, it has implications beyond your log module.

I'd instantiate your own logger instead to dump to a file, instead of tweaking `logging` module level config: https://docs.python.org/2.7/howto/logging.html#configuring-logging

server/lib/python/cartodb_services/requirements.txt
```diff
@@ -2,6 +2,7 @@ redis==2.10.5
 hiredis==0.1.5
 python-dateutil==2.2
 googlemaps==2.4.2
+rollbar==0.13.2
```
rafatower: Having both `requirements.txt` and a `setup.py` with dependencies is kind of a contradiciton. I would favor the `setup.py`, move all the requirements there and remove the other.

Is this correctly automated with chef or is it done manually?

ethervoid: requirements.txt is a default element of python. setup.py is por pip and don't have requirements. This requirements is informative, the package install should be change in chef recipe along with the config part

rafatower: of course I might be wrong XD

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
rafatower: why is the default environment right now if the property is missing?

ethervoid: I've make a change and is development

server/extension/sql/100_routing_helper.sql
```diff
@@ -22,7 +22,7 @@ RETURNS cdb_dataservices_server.simple_route AS $$
   redis_conn = GD["redis_connection_{0}".format(username)]['redis_metrics_connection']
   user_routing_config = GD["user_routing_config_{0}".format(username)]
 
-  logger = Logger(user_routing_config)
+  logger = Logger(plpy)
```
rafatower: what I meant is that it should receive a logger_config object instead, shouldn't it?

server/lib/python/cartodb_services/cartodb_services/tools/log.py
```diff
@@ -0,0 +1,175 @@
...
+import logging
+import json
+import traceback
+import sys
+# Monkey path because plpython sys module doesn't have argv and rollbar
+# package use it
+sys.__dict__['argv'] = []
+
+
+class Logger:
```
rafatower: nice to have: are there any tests for this class?

server/extension/sql/100_routing_helper.sql
ethervoid: I wouldn't do that because we need to create a new object above the logger that don't provides anything than problems because if it changes in the future we have to change all the SQL. The regular way is to create a config object but in extension i thinks is cumbersome

server/lib/python/cartodb_services/cartodb_services/tools/log.py
ethervoid: Nope but I'm thinking if needed because all are third party "services"

server/lib/python/cartodb_services/cartodb_services/tools/log.py
ethervoid: I should not have problems tweaking the logging module, the getLogger method is used to difference between modules but we're using one logger. But I'll change it to have a more regular code

server/extension/sql/100_routing_helper.sql
rafatower: It is just using a factory for LoggerDBConfig or directly to the Logger if you prefer it. as the code stands now

ethervoid: Changed, now it uses LoggerConfig that is passed in the SQL functions

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
rafatower: I don't understand this.

rafatower:
:+1:

server/extension/sql/70_ips.sql
```diff
@@ -20,12 +23,10 @@ RETURNS Geometry AS $$
...
         return None
     except BaseException as e:
-      import sys, traceback
-      type_, value_, traceback_ = sys.exc_info()
+      import sys
       quota_service.increment_failed_service_use()
-      error_msg = 'There was an error trying to geocode using admin0 geocoder: {0}'.format(e)
-      plpy.notice(traceback.format_tb(traceback_))
-      plpy.error(error_msg)
+      logger.error('Error trying to geocode postal code polygon', sys.exc_info(), data={"username": username, "orgname": orgname})
```
iriberri: This should be IP Address

server/extension/sql/80_isolines_helper.sql
```diff
@@ -44,12 +48,10 @@ RETURNS SETOF cdb_dataservices_server.isoline AS $$
...
       return []
   except BaseException as e:
-    import sys, traceback
-    type_, value_, traceback_ = sys.exc_info()
+    import sys
     quota_service.increment_failed_service_use()
-    error_msg = 'There was an error trying to obtain isodistances using here maps geocoder: {0}'.format(e)
-    plpy.notice(traceback.format_tb(traceback_))
-    plpy.error(error_msg)
+    logger.error('Error trying to get mapzen isolines', sys.exc_info(), data={"username": username, "orgname": orgname})
```
iriberri: The client here is HERE, isn't it?

server/lib/python/cartodb_services/cartodb_services/mapzen/matrix_client.py
```diff
@@ -40,6 +42,16 @@ def one_to_many(self, locations, costing):
         }
         response = requests.get(self.ONE_TO_MANY_URL, params=request_params)
 
-        response.raise_for_status() # raise exception if not 200 OK
+        raise Exception('testing exception handling')
```
iriberri: ?

server/lib/python/cartodb_services/cartodb_services/mapzen/routing.py
```diff
@@ -48,7 +49,11 @@ def calculate_route_point_to_point(self, waypoints, mode,
         elif response.status_code == requests.codes.bad_request:
             return MapzenRoutingResponse(None, None, None)
         else:
-            response.raise_for_status()
+            self._logger.error('Error trying to calculate route using HERE',
```
iriberri: It's Mapzen routing

PR #250, opened 2016-08-04
Change notice for info in the logger method

ethervoid:
- Fixed error when to return an empty route because logger object doesn't have `notice` as a method.


ethervoid:

rafatower:
:+1:

PR #263, opened 2016-08-24
Add path to server_conn_str functions

iriberri:
To be released in the next release, unless we edit these functions completely in any other branch. :)


iriberri:
I'm going to move these changes to the `augment_revamp_analysis` branch.

PR #265, opened 2016-08-26
Mapzen to geocode city names

ethervoid:
- Now named place functions uses Mapzen as the first option and in case of any error the fallback is to our internal geocoder


ethervoid:
@iriberri please CR this :)

server/extension/test/sql/00_install_test.sql
```diff
@@ -4,6 +4,7 @@ CREATE EXTENSION plpythonu;
 CREATE EXTENSION plproxy;
 CREATE EXTENSION cartodb;
 CREATE EXTENSION cdb_geocoder;
+CREATE EXTENSION postgres_fdw;
```
iriberri: Why would we need `postgres_fdw` in this extension?

server/lib/python/cartodb_services/cartodb_services/mapzen/geocoder.py
```diff
@@ -47,13 +47,14 @@ def geocode(self, searchtext, city=None, state_province=None, country=None):
...
-                                   state_province=None, country=None):
+                                   state_province=None, country=None,
+                                   search_type=None):
         request_params = {}
         search_string = self._build_search_text(searchtext.strip(),
                                                 city,
                                                 state_province)
         request_params['text'] = search_string
-        request_params['layers'] = 'address'
+        request_params['layers'] = search_type if search_type else 'address'
```
iriberri: I understand this was only used for street geocoding functions before, but out of the scope of this PR I think this line (or defaulting to 'address') might be confusing. Just an opinion though.

server/extension/sql/15_config_helper.sql
```diff
@@ -19,7 +19,7 @@ RETURNS boolean AS $$
     from cartodb_services.metrics import GeocoderConfig
     plpy.execute("SELECT cdb_dataservices_server._connect_to_redis('{0}')".format(username))
     redis_conn = GD["redis_connection_{0}".format(username)]['redis_metadata_connection']
-    geocoder_config = GeocoderConfig(redis_conn, plpy, username, orgname)
+    geocoder_config = GeocoderConfig(redis_conn, plpy, username, orgname, provider)
```
iriberri: If the `geocoder_config` is going to be different depending on the provider with which this function is called, does it make sense to add it to the GD with the same key for all of them? (`"user_geocoder_config_{username}`). In other pieces of code we're using slightly different keys, more specific to the config type, such as `"user_internal_geocoder_config_{username}`.

iriberri: If the `geocoder_config` is going to be different depending on the provider with which this function is called, does it make sense to add it to the GD with the same key for all of them? (`"user_geocoder_config_{username}`). In other pieces of code we're using slightly different keys, more specific to the config type, such as `"user_internal_geocoder_config_{username}`.

server/extension/test/sql/00_install_test.sql
ethervoid: Because we're installing observatory that uses it.

iriberri: Oh, okay, I see the point as we added the dependency there. Then this is out of the scope but I wonder why are we installing observatory here. The server tests are just checking that the functions created in this API exist, but we are not calling to any observatory functions.

server/extension/sql/15_config_helper.sql
ethervoid: I don't want to differentiate from providers. The provider we don't have inside the GeocoderConfig, because it behaves differently, is the internal one. 

What I've just added is a way to force the provider to be used despite the configuration values the user has.

iriberri: :+1:

iriberri:
LGTM :)

server/extension/test/sql/00_install_test.sql
ethervoid: You're right. I don't know why we had this in the installation tests. Removed

server/lib/python/cartodb_services/cartodb_services/mapzen/geocoder.py
ethervoid: Now if you don't pass the search_type, we don't filter by type of search. We use all the possible options.

iriberri: This is cool! 🎊

PR #267, opened 2016-08-30
Table-level OBS_GetMeasue revamp

iriberri:
Restructured to fit better within the camshaft workflow.

It's mostly what we have in master now but:
* Better python format
* Better SQL format
* GetTable function gets split and renamed
* I use INSERT INTOs instead of CREATE TABLEs
* Updated tests for client and server
* function name is no longer a param due to this being exposed in the actual function names

### DIFF
--- a/client/sql/20_table_augmentation.sql
+++ b/client/sql/20_table_augmentation.sql
@@ -111,9 +111,12 @@ CREATE OR REPLACE FUNCTION cdb_dataservices_client.__DST_PrepareTableOBS_GetMeas
             params=plpy.quote_literal(params)
             )
         )
+    if ds_return_metadata[0]["colnames"]:
+        colnames_arr = ds_return_metadata[0]["colnames"]
+        coltypes_arr = ds_return_metadata[0]["coltypes"]
+    else:
+        raise Exception('Error retrieving OBS_GetMeasure metadata')
 
-    colnames_arr = ds_return_metadata[0]["colnames"]
-    coltypes_arr = ds_return_metadata[0]["coltypes"]
 
     # Prepare column and type strings required in the SQL queries
     columns_with_types_arr = [colnames_arr[i] + ' ' + coltypes_arr[i] for i in range(0,len(colnames_arr))]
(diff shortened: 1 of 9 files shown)


iriberri:
Hi @ethervoid!

client/sql/20_table_augmentation.sql
```diff
@@ -1,13 +1,15 @@
 CREATE TYPE cdb_dataservices_client.ds_fdw_metadata as (schemaname text, tabname text, servername text);
 CREATE TYPE cdb_dataservices_client.ds_return_metadata as (colnames text[], coltypes text[]);
 
-CREATE OR REPLACE FUNCTION cdb_dataservices_client._OBS_GetTable(table_name text, output_table_name text, function_name text, params json)
-RETURNS boolean AS $$
+CREATE OR REPLACE FUNCTION cdb_dataservices_client._DST_PrepareTableOBS_GetMeasure(
+    output_table_name text,
+    function_name text,
```
iriberri: Actually we wouldn't need this argument if we're using explicit signatures per each function.

client/sql/20_table_augmentation.sql
```diff
@@ -68,201 +76,191 @@ BEGIN
...
+        "FROM cdb_dataservices_client._DST_ConnectUserTable({username}::text, {orgname}::text, {user_db_role}::text, "
+        "{schema}::text, {dbname}::text, {table_name}::text);" .format(
+            username=plpy.quote_nullable(username),
+            orgname=plpy.quote_nullable(orgname),
+            user_db_role=plpy.quote_literal(user_db_role),
+            schema=plpy.quote_literal(user_schema),
+            dbname=plpy.quote_literal(dbname),
+            table_name=plpy.quote_literal(table_name)))
+
+    server_schema = ds_fdw_metadata[0]["schemaname"]
```
ethervoid: I'll check for empty elements in order to return a more friendly exception

ethervoid:
LGTM

iriberri:
@ethervoid just checked the server return values just in case they're missing. 

Tested them too:

````
cartodb_dev_user_e34eb5ff-70ac-4074-8447-17c981a6f2b2_db=> select _DST_PopulateTableOBS_GetMeasure('prueba_dst', 'tabla8883','{"measure_id":["total_pop"]}'::json);
ERROR:  Exception: Error connecting dataset via FDW
````

ethervoid:
Cool! 💃

PR #270, opened 2016-09-05
Mapzen isochrones integrated

ethervoid:
- Changed cdb_ischrones to use the [new isochrones API](https://mapzen.com/documentation/mobility/isochrone/api-reference/) released by Mapzen


server/extension/cdb_dataservices_server.control
```diff
@@ -1,5 +1,5 @@
 comment = 'CartoDB dataservices server extension'
-default_version = '0.15.1'
+default_version = '0.16'
```
rafatower: not to be picky but.. `0.16.0`?

rafatower:
made a couple of minor comments, other than that looks nice!

server/extension/cdb_dataservices_server.control
ethervoid: Yep, you're right :)

ethervoid:
Maybe we could use it as our Mapzen isochrone function in productions

// @javisantana

javisantana:
could you ask them?

ethervoid:
Sure, I'll let them know to check if they have any inconveniences to have this in production

ethervoid:
Merged and checked against the API they published for the stable version [here](https://mapzen.com/documentation/mobility/isochrone/api-reference/)

There are two new options called `denoise` and `generalize` that are not implemented util we value them as useful :)

ethervoid:
Please @rafatower could you check this please :)

server/extension/sql/80_isolines_helper.sql
```diff
@@ -57,11 +57,9 @@ RETURNS SETOF cdb_dataservices_server.isoline AS $$
     quota_service.increment_total_service_use()
 $$ LANGUAGE plpythonu SECURITY DEFINER;
 
-
 CREATE OR REPLACE FUNCTION cdb_dataservices_server._cdb_mapzen_isolines(
```
rafatower: If I understand it correctly, I'd rename this to `_cdb_mapzen_isodistances`

rafatower:
just a minor change request

Issue #273, opened 2016-09-15
Refactor of configurations and quota stuff: decouple from redis


rafatower:
In order to decouple the dataservices server from redis, the following is required:
- [x] environment in cdb_conf in dataservices server DB
- [x] Factory for redis_conn (this is mostly cleanup of legacy code)
# 
- [x] ConfigInterface: make current config code implement that interface and add a factory for it.
- [ ] add class LocalConfig (or similar) and make the factory instantiate one or another depending on environment.
# 
- [ ] Apply same approach for QuotaService: interface, factory and LocalQuota

@javisantana

PR #274, opened 2016-09-15
Redis refactor

iriberri:
#273


rafatower:
What's here is a big design change:
- move down the stack all low level concerns: "db connections" and redis connections
- avoid too much global state scattered through `GD` var
- encapsulate the redis connection creation in a factory

The whole thing so far is working E2E for mapzen geocoder. We'd like you to take a look before it goes any further, cause this refactor affects the whole server, before arch details are hidden in a pile of commits.

server/lib/python/cartodb_services/cartodb_services/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+#     import cartodb_services
+#     cartodb_services.init(GD)
+#
+#     # rest of the code here
+#
+# $$ LANGUAGE plpythonu;
+
+def init(GD):
+    import plpy
+    plpy.GD = GD
```
rafatower: IMHO GD should be part of `plpy` module and this is a workaround for that design flaw in plpythonu language.

server/lib/python/cartodb_services/cartodb_services/config/environment.py
```diff
@@ -0,0 +1,20 @@
+from db_config import DBConfig
+import json
+
+class Environment:
+    def __init__(self):
+        self._db_config = DBConfig()
```
ethervoid: This should be a dependency to test it properly

server/lib/python/cartodb_services/cartodb_services/metrics/user.py
```diff
@@ -10,9 +11,9 @@ class UserMetricsService:
     SERVICE_HERE_ISOLINES = 'here_isolines'
     DAY_OF_MONTH_ZERO_PADDED = '%d'
 
-    def __init__(self, user_geocoder_config, redis_connection):
+    def __init__(self, user_geocoder_config):
         self._user_geocoder_config = user_geocoder_config
-        self._redis_connection = redis_connection
+        self._redis_connection = RedisConnectionFactory.get_metrics_connection(user_geocoder_config.username)
```
ethervoid: I assume this is a first iteration, and this is going to be deleted in next iterations

server/extension/sql/20_geocode_street.sql
```diff
@@ -141,14 +142,15 @@ RETURNS Geometry AS $$
   from cartodb_services.mapzen.types import country_to_iso3
   from cartodb_services.metrics import QuotaService
   from cartodb_services.tools import Logger,LoggerConfig
-
-  redis_conn = GD["redis_connection_{0}".format(username)]['redis_metrics_connection']
-  user_geocoder_config = GD["user_geocoder_config_{0}".format(username)]
+  from cartodb_services.tools.redis_tools import RedisConnectionFactory
```
ethervoid: Why is this here? is not used

iriberri: Yep. You're right. This comes from a previous interation where the Redis connection was being instantiated to be sent to QuotaService, but it should take care of getting the connections by itself.

rafatower: you're right, this comes from a previous step in the refactor. Thanks!

server/lib/python/cartodb_services/cartodb_services/config/environment.py
rafatower: alright

server/lib/python/cartodb_services/cartodb_services/metrics/user.py
rafatower: most likely not deleted but modified: abstracted and hidden behind a factory that will return one implementation or another depending on config

as stated in the review comment: this is by no means final, just wanted you to :eyes: into it before going any further as things will be much harder to review in later phases

server/lib/python/cartodb_services/cartodb_services/tools/redis_tools.py
```diff
@@ -1,8 +1,31 @@
...
 import json
+import plpy
 
 
+class RedisConnectionFactory:
+
+    @classmethod
+    def get_metadata_connection(cls, username):
+        cache_key = "redis_connection_{0}_metadata".format(username)
+        if cache_key in plpy.GD:
```
ethervoid: What I mean, is that with the current version we cache the connection and the config (for example GeocoderConfig) because config doesn't change so we avoid to query Redis every time to get the config.

rafatower: Point taken. That said I will not mind about (early) optimizations until they prove to be a problem. Anyway caching is an implementation detail that will be isolated rather than scattered in `GD` usage

rafatower:
@ethervoid & @jgoizueta the code WIP in case you want to take a look. I'll shuffle things around for sure, particularly object instantiation

server/extension/sql/20_geocode_street.sql
```diff
@@ -165,16 +165,16 @@ RETURNS Geometry AS $$
   configs = request.configs
 
   # TODO change to hires_geocoder_config = MapzenGeocoderConfigFactory.get(request)
-  user_geocoder_config = MapzenGeocoderConfigFactory(configs).get(user)
+  mapzen_geocoder_config = MapzenGeocoderConfigFactory(configs).get(user)
 
-
-  quota_service = QuotaService(user_geocoder_config)
+  from cartodb_services.refactor.quota.mapzen_quota_service import MapzenGeocoderQuotaService
```
iriberri: I'm missing the actual module restructuration with this commit, but I guess it's coming! ✨

rafatower: Actually I prefer to work that into another branch and maybe close this pull request. I'll brief you on the details later on.

iriberri:
I'm closing this PR, we're moving to branch `redis-refactor-take2`. Please don't delete this branch!

PR #275, opened 2016-09-16
Fixed QPS retry decorator

ethervoid:
Until now we weren't using the QPS retry because the response object was always `None` and we couldn't check if the error code was 429.

That's is fixed now. I've added optional parameters to the decorator in order to make possible to tweak the conditions for every case

// @jgoizueta please review this :)

### DIFF
--- a/server/lib/python/cartodb_services/cartodb_services/mapzen/qps.py
+++ b/server/lib/python/cartodb_services/cartodb_services/mapzen/qps.py
@@ -4,18 +4,38 @@ from datetime import datetime
 from exceptions import TimeoutException
 
 DEFAULT_RETRY_TIMEOUT = 60
+DEFAULT_QUERIES_PER_SECOND = 10
 
-
-def qps_retry(f):
-    def wrapped_f(*args, **kw):
-        return QPSService().call(f, *args, **kw)
-    return wrapped_f
+def qps_retry(original_function=None,**options):
+    """ Query Per Second retry decorator
+        The intention of this decorator is to retry requests against third
+        party services that has QPS restriction.
+        Parameters:
+            - timeout: Maximum number of seconds to retry
+            - qps: Allowed queries per second. This parameter is used to
+                   calculate the next time to retry the request
+    """
+    if original_function is not None:
+        def wrapped_function(*args, **kwargs):
+            if 'timeout' in options:
+                timeout = options['timeout']
+            else:
+                timeout = DEFAULT_RETRY_TIMEOUT
+            if 'qps' in options:
+                qps = options['qps']
+            else:
+                qps = DEFAULT_QUERIES_PER_SECOND
+            return QPSService(retry_timeout=timeout, queries_per_second=qps).call(original_function, *args, **kwargs)
+        return wrapped_function
+    else:
+        def partial_wrapper(func):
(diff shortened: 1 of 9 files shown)


server/lib/python/cartodb_services/cartodb_services/mapzen/qps.py
```diff
@@ -4,18 +4,38 @@
...
+    """ Query Per Second retry decorator
+        The intention of this decorator is to retry requests against third
+        party services that has QPS restriction.
+        Parameters:
+            - timeout: Maximum number of seconds to retry
+            - qps: Allowed queries per second. This parameter is used to
+                   calculate the next time to retry the request
+    """
+    if original_function is not None:
+        def wrapped_function(*args, **kwargs):
```
jgoizueta: How/when occurs this? (!original_function)

jgoizueta: Correction: I meant to comment the other branch (when original_function is None)

ethervoid:
// @rafatower @iriberri

server/lib/python/cartodb_services/test/test_qps.py
```diff
@@ -0,0 +1,33 @@
...
+
+requests_mock.Mocker.TEST_PREFIX = 'test_'
+
+@requests_mock.Mocker()
+class TestQPS(TestCase):
+    QPS_ERROR_MESSAGE = "Queries per second exceeded: Queries exceeded (10 allowed)"
+
+    def test_qps_timeout(self, req_mock):
+        class TestClass:
+            @qps_retry(timeout=1)
```
rafatower: This test takes 3 seconds vs the typical 100ms of the whole suite.

server/lib/python/cartodb_services/test/test_qps.py
```diff
@@ -0,0 +1,33 @@
...
+                if response.status_code == 429:
+                    raise ServiceException('Error 429', response)
+
+        def _text_cb(request, context):
+            context.status_code = 429
+            return self.QPS_ERROR_MESSAGE
+
+        req_mock.register_uri('GET', 'http://localhost/test_qps',
+                       text=_text_cb)
+        with self.assertRaises(TimeoutException):
```
rafatower: To check the actual behavior a check on the number of requests or the time elapsed between requests would be needed here.

I'm gonna try to improve it by mocking the sleep... if you have a better suggestion I'm all hears.

server/lib/python/cartodb_services/test/test_qps.py
ethervoid: Yep, because try to execute it until it ends. Just one slow test to check the QPS retry and raise timeout. No big deal

server/lib/python/cartodb_services/test/test_qps.py
ethervoid: Up to you :)

rafatower:
After messing around for a while and trying to mock things I tried the most cost-effective approach: raising the QPS and lowering the timeout:
```
@qps_retry(timeout=0.001, qps=1000)
```

with those I'm still getting execution times above 1 second, so I don't think this is ready.

rafatower:
fixed

PR #278, opened 2016-09-20
Redis refactor user config

iriberri:


iriberri:
@rafatower FYI, this is what I have so far. Pushed it to a separate branch so that it's easier to read and review, as we were doing with other changes.

server/lib/python/cartodb_services/cartodb_services/config/user_config.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def get(self):
+        if self._user_config_obj:
+            return self._user_config_obj
+        elif self._environment = 'onpremises':
+            # Asuming there's no separate organization settings
+            # because it is all set in server DB
+            return DbUserConfig()
+        elif self._orgname is not None:
+            return RedisOrgConfig(orgname)
```
rafatower: to my understanding here is the hairy detail that I'm not sure about yet.

rafatower:
Looks like a good starting point to me, let's further discuss this tomorrow.

PR #284, opened 2016-09-28
Redis refactor: Take 2

iriberri:


rafatower:
Hey @jgoizueta and @ethervoid I got this review for you. This one must be a lot more cleaner than the previous one from a reviewer's standpoint.

What this does:
- it creates a new "namespace" `refactor`, and 99% of the new code is kept there.
- there are tree main modules:
    - `storage`: low level details, regarding db, redis and more (notably in-memory and null storages)
    - `backend`: this layer abstracts over the actual storages so that the upper layers don't know about those details. In general the factories build backends and the client classes only need to know that they implement a common interface.
   - `service`: this is actually the one that contains most of the business logic.
- It re-implements configs using "value objects" and builders to generate them.
- A Config used by a particular "service" only contain what a particular service needs, following the "Law of Demeter"

My goal no. 1 with the review is to detect weak spots I might not be aware of and to expedite deploy whenever possible, then extend it to the whole geocoder file and to other services.

PR #291, opened 2016-10-18
281 quota mapzen routing

rafatower:
@jgoizueta can you please take a look? The main point here is to make sure redis keys match in both sides of the equation.

~~(There's a disturbing lack of tests. I know, but the existing ones run as expected :smile: . I'll write some tests for this stuff once we agree on this piece of code)~~ with tests now.

### DIFF
--- a/server/lib/python/cartodb_services/cartodb_services/metrics/config.py
+++ b/server/lib/python/cartodb_services/cartodb_services/metrics/config.py
@@ -156,7 +156,7 @@ class RoutingConfig(ServiceConfig):
         self._monthly_quota = self._get_effective_monthly_quota()
 
     def _get_effective_monthly_quota(self):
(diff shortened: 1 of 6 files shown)


jgoizueta:
Don't we need to add the routing service [here](https://github.com/CartoDB/dataservices-api/blob/development/server/lib/python/cartodb_services/cartodb_services/metrics/user.py#L19-L23)?

rafatower:
You are 100% right. Actually defaulting to the geocoder quota is not the best option IMHO.

rafatower:
Now with tests:
```
ubuntu@cdb:~/src/dataservices-api/server/lib/python/cartodb_services$ nosetests  test/
.....................................................................................................
----------------------------------------------------------------------
Ran 101 tests in 0.124s

OK
```

@jgoizueta last peek into the code? I'm going to test end-to-end locally and in staging but I'm pretty confident with the code now.

PR #292, opened 2016-10-19
fixed broken hyperlink

csobier:
Fixed one broken hyperlink that broken when the rebranding baseurl was applied.

@ethervoid, can I merge this fix in Dev and fix directly in Master too?


ethervoid:
Merged into master, now I'll rebase development with master.

Issue #293, opened 2016-10-26
When mapzen returns empty in one-to-many service we fail


ethervoid:
If the response doesn't have JSON in the body, we fail to parse it:

```
Traceback (most recent call last):
1
File "<string>" line 38 in __plpython_procedure__cdb_mapzen_isolines_829328 [locals]
2
File "/usr/local/lib/python2.7/dist-packages/cartodb_services/mapzen/isolines.py" line 43 in calculate_isochrone [args] [locals]
return self.calculate_isoline(origin, costing_model, time_range, upper_rmax, 'time')
3
File "/usr/local/lib/python2.7/dist-packages/cartodb_services/mapzen/isolines.py" line 106 in calculate_isoline [args] [locals]
response = self._matrix_client.one_to_many([origin] + location_estimates,  costing_model)
4
File "/usr/local/lib/python2.7/dist-packages/cartodb_services/mapzen/matrix_client.py" line 56 in one_to_many [args] [locals]
return response.json()
5
File "/usr/local/lib/python2.7/dist-packages/requests/models.py" line 808 in json [args] [locals]
return complexjson.loads(self.text, **kwargs)
6
File "/usr/lib/python2.7/dist-packages/simplejson/__init__.py" line 413 in loads [args] [locals]
return _default_decoder.decode(s)
7
File "/usr/lib/python2.7/dist-packages/simplejson/decoder.py" line 402 in decode [args] [locals]
obj, end = self.raw_decode(s, idx=_w(s, 0).end())
8
File "/usr/lib/python2.7/dist-packages/simplejson/decoder.py" line 420 in raw_decode [args] [locals]
raise JSONDecodeError("No JSON object could be decoded", s, idx)
JSONDecodeError: No JSON object could be decoded: line 1 column 0 (char 0)
```

PR #294, opened 2016-10-26
Fix empty response from matrix

ethervoid:
Closes https://github.com/CartoDB/dataservices-api/issues/293

### DIFF
--- a/server/lib/python/cartodb_services/cartodb_services/mapzen/isolines.py
+++ b/server/lib/python/cartodb_services/cartodb_services/mapzen/isolines.py
@@ -86,7 +86,7 @@ class MapzenIsolines:
     def calculate_isoline(self, origin, costing_model, isorange, upper_rmax, cost_variable, unit_factor=1.0):
 
         # NOTE: not for production
-        self._logger.debug('Calculate isoline', data={"origin": origin, "costing_model": costing_model, "isorange": isorange})
+        # self._logger.debug('Calculate isoline', data={"origin": origin, "costing_model": costing_model, "isorange": isorange})
 
         # Formally, a solution is an array of {angle, radius, lat, lon, cost} with cardinality NUMBER_OF_ANGLES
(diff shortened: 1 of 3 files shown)


ethervoid:
Please @rafatower check this out :)

server/lib/python/cartodb_services/cartodb_services/mapzen/isolines.py
```diff
@@ -105,14 +105,16 @@ def calculate_isoline(self, origin, costing_model, isorange, upper_rmax, cost_va
 
             response = self._matrix_client.one_to_many([origin] + location_estimates,  costing_model)
             costs = [None] * self.NUMBER_OF_ANGLES
+            if not response:
```
rafatower: I did not now that the negation of an empty `dict` evaluates to `True`:
```python
In [1]: not {}
Out[1]: True
```

I'd test this with some extreme case. I.e: all responses are empty for some reason. What it should do is to stick to the original estimate. Maybe you can test it by asking for isolines from a point in the middle of the ocean.

server/lib/python/cartodb_services/cartodb_services/mapzen/matrix_client.py
```diff
@@ -52,6 +52,16 @@ def one_to_many(self, locations, costing):
...
+            if response.status_code == requests.codes.bad_request:
+                return {}
+            else:
+                raise ServiceException("Error trying to get matrix distance from mapzen", response)
 
-        return response.json()
+        # response could return with empty json
+        try:
+            return response.json()
+        except:
```
ethervoid: I thought about it, and was my first option but I put this generic `except` just in case.

server/lib/python/cartodb_services/cartodb_services/mapzen/isolines.py
ethervoid: Yes, an empty `dict` acts as falsy.

Ok, let check it out!

PR #299, opened 2016-10-28
Data services metrics enhanced

ethervoid:
- Added metrics context manager to gather data from different parts of the server functions
- Support multiple response data for one server function call: For example in the one_to_many matrix client
- Metrics files configuration is not mandatory
- All the services covered and gathering metrics


server/extension/sql/30_admin0.sql
```diff
@@ -13,23 +14,24 @@ RETURNS Geometry AS $$
...
-      quota_service.increment_empty_service_use()
-      return None
-  except BaseException as e:
-    import sys
-    quota_service.increment_failed_service_use()
-    logger.error('Error trying to geocode admin0 polygon', sys.exc_info(), data={"username": username, "orgname": orgname})
-    raise Exception('Error trying to geocode admin0 polygon')
-  finally:
-    quota_service.increment_total_service_use()
+  with metrics('cdb_geocode_admin0_polygon', user_geocoder_config):
```
rafatower: what is used from the service config, other than the service name?

rafatower: okay, I see it's more for compat with the geocoder metrics.

ethervoid: From service_config I get: username, orgname, service_type, provider...

ethervoid:
Yes, now we have generic for all the services except for Geocoding, in the future, it's very probable to have to specialize some logger and add specific fields.

ethervoid:
Please @rafatower another check now it's working.

server/extension/sql/120_data_observatory_geometries.sql
```diff
@@ -31,22 +32,23 @@ RETURNS geometry(Geometry, 4326) AS $$
...
-        quota_service.increment_empty_service_use()
-        return None
-  except BaseException as e:
-      import sys
-      quota_service.increment_failed_service_use()
-      logger.error('Error trying to OBS_GetBoundary', sys.exc_info(), data={"username": username, "orgname": orgname})
-      raise Exception('Error trying to OBS_GetBoundary')
-  finally:
-      quota_service.increment_total_service_use()
+  with metrics('obs_getboundary', user_obs_snapshot_config, logger):
```
rafatower: why are we measuring in some cases "public" functions and in other cases "private" functions? E.g: `obs_getboundary` and `_OBS_GetAvailableBoundaries`

server/lib/python/cartodb_services/cartodb_services/metrics/log.py
```diff
@@ -2,73 +2,216 @@
...
+    @classmethod
+    def get_element(self, key):
+        return MetricsDataGatherer.instance().get_element(key)
+
+    @classmethod
+    def clean(self):
+        MetricsDataGatherer.instance().clean()
+
+    @classmethod
+    def instance(self):
```
rafatower: is this a singleton? a bit of code doc could help here

server/extension/sql/120_data_observatory_geometries.sql
rafatower: mm they seem to be aggregated by their `function_name`, right?

rafatower: how does it work when nesting function calls? E.g in pseudo-code:

```sql
CREATE OR REPLACE FUNCTION cbd_isochorne(...)
AS $$
    with metrics('cdb_isochrone'...):
        plpy.execute('SELECT _cdb_mapzen_isochrone(...)')
$$ LANGUAGE plpythonu;

CREATE OR REPLACE FUNCTION _cdb_mapzen_isochrone(...)
AS $$
    with metrics('cdb_isochrone'...):
         ...
$$ LANGUAGE plpythonu;
```

probably I'm not getting something ¯\_(ツ)_/¯

server/extension/sql/110_data_observatory_augmentation.sql
```diff
@@ -11,9 +11,9 @@ RETURNS text AS $$
   plpy.execute("SELECT cdb_dataservices_server._connect_to_redis('{0}')".format(username))
   redis_conn = GD["redis_connection_{0}".format(username)]['redis_metrics_connection']
   plpy.execute("SELECT cdb_dataservices_server._get_obs_snapshot_config({0}, {1})".format(plpy.quote_nullable(username), plpy.quote_nullable(orgname)))
-  user_obs_snapshot_config = GD["user_obs_snapshot_config_{0}".format(username)]
+  user_obs_config = GD["user_obs_config_{0}".format(username)]
 
-  return user_obs_snapshot_config.connection_str
+  return user_obs_config.connection_str
```
rafatower: should it be calling `SELECT cdb_dataservices_server._get_obs_config(...)` instead?

server/extension/sql/120_data_observatory_geometries.sql
ethervoid: Should be used in the public functions. There is an exception and is namedplace because breaks the tests because we need to add the config call and try to connect to redis and fails the regress tests. We need to include a mock for the redis part until we finish the refactor

AFAIK in the other functions should be in the public functions, so in that case is an error.

server/lib/python/cartodb_services/cartodb_services/metrics/log.py
ethervoid: Yup is a singleton. Ok I'll add some docstring there

rafatower:
:+1: I'll leave the details to you :)

PR #305, opened 2016-11-10
302 expose quotas

rafatower:
Early review request. Things done here:
- Added functions to check:
  - monthly quota allowance, 
  - remaining quota for the current billing period,
  - soft limit, 
  - provider
- Modified the config and metrics service to return values as integers

Things to do:
- extend to the rest of services
- add the client's part
- add tests
- integrate into analysis pre-checks

@jgoizueta and @ethervoid can you please take a quick look? mostly interested in how you find the API and how it fits into the rest of the codebase.

From the top of my head: best interface to deal with soft limits? maybe just instead of  remaining quota expose the used quota?

### DIFF
--- a/server/extension/sql/200_quotas.sql
+++ b/server/extension/sql/200_quotas.sql
@@ -37,3 +37,20 @@ RETURNS integer AS $$
   else:
     raise 'not implemented'
 $$ LANGUAGE plpythonu;
+
+
+CREATE OR REPLACE FUNCTION cdb_dataservices_server.cdb_soft_limit(
+  username TEXT,
+  orgname TEXT,
+  service TEXT)
+RETURNS boolean AS $$
+  plpy.execute("SELECT cdb_dataservices_server._connect_to_redis('{0}')".format(username))
+  redis_conn = GD["redis_connection_{0}".format(username)]['redis_metrics_connection']
+
+  if service == 'isolines':
+    plpy.execute("SELECT cdb_dataservices_server._get_isolines_routing_config({0}, {1})".format(plpy.quote_nullable(username), plpy.quote_nullable(orgname)))
+    user_isolines_config = GD["user_isolines_routing_config_{0}".format(username)]
+    return user_isolines_config.soft_isolines_limit
+  else:
+    raise 'not implemented'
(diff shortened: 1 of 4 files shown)


server/extension/sql/200_quotas.sql
```diff
@@ -0,0 +1,73 @@
...
+  redis_conn = GD["redis_connection_{0}".format(username)]['redis_metrics_connection']
+
+  if service == 'isolines':
+    plpy.execute("SELECT cdb_dataservices_server._get_isolines_routing_config({0}, {1})".format(plpy.quote_nullable(username), plpy.quote_nullable(orgname)))
+    user_isolines_config = GD["user_isolines_routing_config_{0}".format(username)]
+    monthly_quota = user_isolines_config.isolines_quota
+
+    user_service = UserMetricsService(user_isolines_config, redis_conn)
+    current_used = user_service.used_quota(user_isolines_config.service_type, date.today())
+    return monthly_quota - current_used
```
jgoizueta: In the case of soft limits, can't the used quota be greater than the monthly quota?

rafatower: Yes, it can. Actually this can return negative numbers in that case. Mathematically correct, but to be used as a pre-check then it forces you into checking both the soft_limit and the remaining quota. I don't know if a more abstract operation would be useful. Maybe something like...:
```sql
CREATE OR REPLACE FUNCTION cdb_dataservices_server.cdb_enough_quota(
   username TEXT,
   orgname TEXT,
   service TEXT,
   input_rows integer)
RETURNS boolean AS $$
-- etc...
$$
```
?

jgoizueta: we'd need example uses cases for these functions to decide...

rafatower:
Okay, thanks for the review. What Mario suggested is to have one function to return `(monthly_quota, used_quota, soft_limit, provider)` as a tuple and another one to check if there's enough quota.

We do agree in that the code cries for a refactor and pushing config/quota details further into the python library (factories for them). The gracious point here is to keep this "API" stable regardless of the future refactor(s).

I'm closing this early review, performing the changes and will reopen another review as soon as I have them. Thanks again! :)

PR #310, opened 2016-11-21
Fix typo: replace "isolines" with "routes" in routing api docs

rafatower:
A bit of context: https://carto.com/docs/carto-engine/dataservices-api/routing-functions/

@csobier can you please review? thanks

cc/ @noguerol


csobier:
Looks good!

PR #311, opened 2016-11-24
302 expose quotas

rafatower:
This is ready for review if you don't mind taking a look @ethervoid 

What's done:
- functions to check the quota, both server and client
- removed the `no_params` from the templates (this caused trouble, not needed anymore)
- bug fixes: observatory quota, quotas as integers, mapzen geocoder soft limit as bool

What's remaining:
- versioning stuff (this is a bit of a pain, all the 3 parts to be released (lib, server and client)
- tests?

I'd like to merge and version in `development` branch, if you don't see any major obstacle. I may ask for a bit of help...

cc/ @jgoizueta just fyi


client/renderer/sql-template-renderer
```diff
@@ -49,11 +45,11 @@ class SqlTemplateRenderer
   end
 
   def params
-    @function_signature['params'].reject(&:empty?).map { |p| "#{p['name']}"}.join(', ')
```
ethervoid: Why did you move this join from here to the templates?

client/sql/16_custom_types.sql
```diff
@@ -18,3 +18,21 @@ CREATE TYPE cdb_dataservices_client.obs_meta_denominator AS (denom_id text, deno
...
+-- For quotas and services configuration
+CREATE TYPE cdb_dataservices_client.service_type AS ENUM (
+    'isolines',
+    'hires_geocoder',
+    'routing',
+    'observatory',
+    'internal_geocoder'
+);
+
+CREATE TYPE cdb_dataservices_client.service_params AS (
```
ethervoid: Maybe service_params is not the best name, it sounds like a type you pass as argument not the result of a query for data

server/extension/sql/200_quotas.sql
```diff
@@ -0,0 +1,120 @@
...
+  from datetime import date
+
+  plpy.execute("SELECT cdb_dataservices_server._connect_to_redis('{0}')".format(username))
+  redis_conn = GD["redis_connection_{0}".format(username)]['redis_metrics_connection']
+
+  today = date.today()
+  ret = []
+
+  #-- Isolines
+  service = 'isolines'
```
ethervoid: We should, in the future refactor, move this to the python library. The main reason it the future changes/addition of services we are going to deal with. It's better to change and deploy the python library than the whole extension.

server/extension/sql/200_quotas.sql
```diff
@@ -0,0 +1,120 @@
...
+
+  monthly_quota = user_routing_config.monthly_quota
+  used_quota = user_service.used_quota(user_routing_config.service_type, today)
+  soft_limit = user_routing_config.soft_limit
+  provider = user_routing_config.provider
+  ret += [[service, monthly_quota, used_quota, soft_limit, provider]]
+
+  #-- Observatory
+  service = 'observatory'
+  plpy.execute("SELECT cdb_dataservices_server._get_obs_config({0}, {1})".format(plpy.quote_nullable(username), plpy.quote_nullable(orgname)))
```
ethervoid: As a remainder. We should check for the snapshot part and merge/remove if it's not used.

server/extension/sql/200_quotas.sql
```diff
@@ -0,0 +1,120 @@
...
+  orgname TEXT,
+  service_ cdb_dataservices_server.service_type,
+  input_size NUMERIC)
+returns BOOLEAN AS $$
+  DECLARE
+    params cdb_dataservices_server.service_params;
+  BEGIN
+    SELECT * INTO params
+      FROM cdb_dataservices_server.cdb_service_params(username, orgname) AS p
+      WHERE p.service = service_;
```
ethervoid: Why don't just put service as parameter in the cdb_service_params function?

server/extension/sql/200_quotas.sql
```diff
@@ -0,0 +1,120 @@
...
+    RETURN params.soft_limit
+           OR params.monthly_quota IS NULL -- account for the internal_geocoder
```
ethervoid: maybe is better to check for 'internal_geocoder'. A bug in the monthly_quota parameter could lead to this returning true

server/lib/python/cartodb_services/cartodb_services/metrics/user.py
```diff
@@ -22,6 +23,8 @@ def used_quota(self, service_type, date):
             return self.__used_isolines_quota(service_type, date)
         elif service_type == self.SERVICE_MAPZEN_ROUTING:
             return self.__used_routing_quota(service_type, date)
+        elif service_type == self.SERVICE_OBSERVATORY:
```
ethervoid: As a remainder. We should check for all the types with regexp like [here](https://github.com/CartoDB/dataservices-api/blob/development/server/lib/python/cartodb_services/cartodb_services/metrics/quota.py#L70-L88)

client/renderer/sql-template-renderer
rafatower: to let the templates decide how to render to strings and get rid of the `no_params` that makes it harder, for the case of `no_params && multi_row`

client/sql/16_custom_types.sql
rafatower: how about `service_quota_info`? any other better name?

server/lib/python/cartodb_services/cartodb_services/metrics/user.py
rafatower: Sometimes I confuse service_provider or provider_service and writing regexps on top of that standard scares me a little XD (you know I'm a gominola)

client/sql/16_custom_types.sql
ethervoid: service_quota_info fits better

server/lib/python/cartodb_services/cartodb_services/metrics/user.py
ethervoid: xDDD

PR #315, opened 2016-12-02
Make the HTTP requests retry on timeout

ethervoid:


rafatower:
Ok to the changes. I'm taking over this and doing the deploy...

Issue #317, opened 2016-12-07
Retries not working as expected


rafatower:
## Description of the problem

The issue is that some services are returning error responses sometimes and the `@qps_retry` is not working as expected. From the error log (see R/B 117):
```python
Traceback (most recent call last):
1 File "<string>" line 31 in __plpython_procedure__cdb_mapzen_isochrones_25880422 [locals]
2 File "/usr/local/lib/python2.7/dist-packages/cartodb_services/mapzen/qps.py" line 28 in wrapped_function [locals]
  'return QPSService(retry_timeout=timeout, querie...nd=qps).call(original_function, *args, **kwargs)'
3 File "/usr/local/lib/python2.7/dist-packages/cartodb_services/mapzen/qps.py" line 51 in call [args] [locals]
  self.retry(start_time, attempt_number)
4 File "/usr/local/lib/python2.7/dist-packages/cartodb_services/mapzen/qps.py" line 59 in retry [args] [locals]
  raise TimeoutException()
TimeoutException: 'Timeout requesting to mapzen server'
```

By inspecting the local variables, turns out that `retry_count	== 1`.

Taking a closer look to the code: https://github.com/CartoDB/dataservices-api/blob/a977bc97ab9add0040c66f1e3cdeaf61dff919e8/server/lib/python/cartodb_services/cartodb_services/mapzen/qps.py#L58

the problem there is that we're comparing the elapsed time in microseconds with the timeout in miliseconds. Using the default timeout, it will indeed fail with a `TimeoutException` if the first request takes more than `60 ms`.

## Tasks
- [x] Fix the QPS logic - DONE in #318
- [x] Discuss the PR #315 and the need of python `requests` lib retries cc/ @ethervoid

ethervoid:
About the `requests` retries are the cleanest way to retry timeouts that we're suffering for. From the requests documentation:
```
Note, this applies only to failed DNS lookups, socket connections and connection timeouts, never to requests where data has made it to the server
```
This change is not related to the QPS part, they're different measures for different problems

rafatower:
:+1: Then it makes all the sense. I was wondering if that could cause a burst of requests faster than the allowed QPS. Thanks for pointing out

rafatower:
So, for consistency I'd revert this line: https://github.com/CartoDB/dataservices-api/pull/318/files#diff-82b4d24735e8e5a9f7fa4bf44eaef74bR16

test and release a new version. cc/ @jgoizueta

[...]

PR #320, opened 2016-12-12
Revert number of isochrone retries to 3

jgoizueta:
See #317 
Note that finally the number of retries has been set at 1 (see below)

### DIFF
--- a/server/lib/python/cartodb_services/cartodb_services/here/geocoder.py
+++ b/server/lib/python/cartodb_services/cartodb_services/here/geocoder.py
@@ -18,7 +18,7 @@ class HereMapsGeocoder(Traceable):
     DEFAULT_GEN = 9
     READ_TIMEOUT = 60
     CONNECT_TIMEOUT = 10
-    MAX_RETRIES=3
+    MAX_RETRIES=0
 
     ADDRESS_PARAMS = [
         'city',
@@ -92,7 +92,7 @@ class HereMapsGeocoder(Traceable):
         request_params.update(params)
         # TODO Extract HTTP client wrapper
         session = requests.Session()
-        session.mount(self.host, HTTPAdapter(self.MAX_RETRIES))
+        session.mount(self.host, HTTPAdapter(max_retries=self.MAX_RETRIES))
         response = session.get(self.host, params=request_params,
                                 timeout=(self.CONNECT_TIMEOUT, self.READ_TIMEOUT))
         self.add_response_data(response, self._logger)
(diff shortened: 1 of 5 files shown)


jgoizueta:
@rafatower  please review

rafatower:
:+1:

rafatower:
as commented in person: let's do the versioning stuff in the developmet branch

rafatower:
:+1:

jgoizueta:
It seems that we were passing the `MAX_RETRIES` parameter as the first argument to the requests' `HTTPAdapter` constructor, which is the connections pool size actually, so it wasn't having the intended effect and the default value of 0 retries was being used.

I've fixed the argument passing part (using a keyword parameter) here: https://github.com/CartoDB/dataservices-api/pull/320/commits/79fb7961802a15effdad772861d94350808793e7
but changing all our `MAX_RETRIES` constants to 0, so that we don't change current behaviour.

I think we should further discuss/test the effects of changing `MAX_RETRIES` since it will affect the effective time till a timeout exception is generated by the retries library, and catched by our QPS stuff, etc.

cc @rafatower @ethervoid

rafatower:
As per Mario's comment in the ticket, that `MAX_RETRIES` affects this:
> Note, this applies only to failed DNS lookups, socket connections and connection timeouts, never to requests where data has made it to the server

The QPS param is in app layer so to say, not network layer. From our side it is used to calculate the delay between requests (exponential backoff) for "app retries" so if we are not 100% accurate with it, I don't think it is big deal. The timeout is by default in a generous amount in seconds.

I wouldn't strive for perfection. What is your concern exactly? what course do you propose?

jgoizueta:
As per our conversation we agree to set the number of (low level) retries to 1.

Summary of our discussion: The low level or network level retries that `requests` handle and are affected by this parameter don't occur for the 429 responses that QPS handles (with exponential backoff retries), so both retry levels (requests/network level vs QPS/app level) are mostly independent.

Having the network level retries will save some cases we're having now of spurious external service timeouts. The only negative effect is that when the external service is down we'll block trying to connect for T*(N+1) seconds where T is the network level [connect timeout](https://github.com/CartoDB/dataservices-api/blob/f716fbb5020e58095a359ea0f98065a9ffcd9dc0/server/lib/python/cartodb_services/cartodb_services/mapzen/geocoder.py#L17) (10s currently) and N the maximum number of retries. So, we'll set N=1 at the moment to fix spurious service problems while being conservative with the maximum blocking time.

PR #323, opened 2016-12-13
314 exception safe public funcs

rafatower:
@jgoizueta @ethervoid  early and short review please.

Some remarks:
- I don't want to open a full debate as we already discussed this in a meeting. But... (keep reading)

- I'm pretty reluctant to adding these as "public functions", especially documentation and API wise (you know, upgrades, downgrades, keeping the same stable interface... in many schools of thought exceptions are part of the API). How about prepending an underscore and using them just from the analysis layer?

- I guess we can add specific treatment to some exceptions here, but IMO that misses the point of the whole thing. There are a few sources of exceptions:
  - programming errors (e.g: observatory raising an exception for a given input) => need to be fixed and  ideally should be detected before reaching prod
  - network issues: already working on retries and all that.
  - 3rd party issues: we can do little with those
  - **quota exhaustion**: this is the most specific case we want to deal with. Shall I deal with just this case??
  - any other I might be missing??

- Adding "unit tests" for this code is fairly easy. Testing E2E is a bit harder, even in local.

- The acceptance IMO is testing E2E with half the quota needed and checking than half the work can be carried out and kept, at least with the 3 types of function interfaces: `multi_row`, `multi_field` and the rest (the "normal" case).

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -57,9 +57,11 @@ all: $(DATA)
 .PHONY: release
 release: $(EXTENSION).control $(SOURCES_DATA)
 	test -n "$(NEW_VERSION)"  # $$NEW_VERSION VARIABLE MISSING. Eg. make release NEW_VERSION=0.x.0
-	mv *.sql old_versions
+	git mv *.sql old_versions
 	$(SED) $(REPLACEMENTS) $(EXTENSION).control
+	git add $(EXTENSION).control
 	cat $(SOURCES_DATA_DIR)/*.sql > $(EXTENSION)--$(NEW_VERSION).sql
+	git add $(EXTENSION)--$(NEW_VERSION).sql
 	$(ERB) version=$(NEW_VERSION) upgrade_downgrade_template.erb > $(EXTENSION)--$(EXTVERSION)--$(NEW_VERSION).sql
 	$(ERB) version=$(EXTVERSION) upgrade_downgrade_template.erb > $(EXTENSION)--$(NEW_VERSION)--$(EXTVERSION).sql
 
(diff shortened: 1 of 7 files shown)


rafatower:
Summary of our in-person debate:
- generic (not only for quota)
- private (we don't want to have it documented and maintained cause potentially it can change or be dropped)
- needs more info in the `WARNING` from the original exception, when possible.

client/Makefile
```diff
@@ -57,9 +57,11 @@ all: $(DATA)
 .PHONY: release
 release: $(EXTENSION).control $(SOURCES_DATA)
 	test -n "$(NEW_VERSION)"  # $$NEW_VERSION VARIABLE MISSING. Eg. make release NEW_VERSION=0.x.0
-	mv *.sql old_versions
+	git mv *.sql old_versions
 	$(SED) $(REPLACEMENTS) $(EXTENSION).control
+	git add $(EXTENSION).control
 	cat $(SOURCES_DATA_DIR)/*.sql > $(EXTENSION)--$(NEW_VERSION).sql
+	git add $(EXTENSION)--$(NEW_VERSION).sql
```
rafatower: What these addition do:
image

Basically it leaves the files in staging area of git, except for the upgrade and downgrade files, so that they can be edited manually, added and committed.

rafatower:
Changes since the last review:
- Made the functions "private" (prepending the underscore and renaming files, etc)
- Added the info that I found useful to the WARNING
- Added "unit" tests to check the behavior for the different types of function interfaces we have.
- Minor addition to the Makefile that I found useful to me.

@ethervoid and @jgoizueta can you please take a look before I version this? Thanks!

Issue #322, opened 2016-12-13
Check quota fails in some cases with period_end_date


ethervoid:
There is a bug with the check quota dates [here](https://github.com/CartoDB/dataservices-api/blob/development/server/lib/python/cartodb_services/cartodb_services/metrics/user.py#L144-L148).

Basically if, for example, we take a period_end_date of `2016-08-31 13:59:49+00:00` that piece of code for today (2016-12-13 13:59:49+00:00) is trying to create a billing cycle from 2016-11-31 to 2016-12-31 which is not possible because November has 30 days.

This leads to `ValueError: day is out of range for month`

As acceptance, we should check and add tests for 30 days months and for February.

jgoizueta:
I'm leaving this notes here about a brief look I took at this yesterday for whoever may need it:

```python
import calendar
def last_day_of_month(y, m):
  return calendar.monthrange(y, m)[1];
```

Then we can use it to construct dates like so:

```python
date(temp_date.year, temp_date.month, min(end_period_day, last_day_of_month(temp_date.year, temp_date.month)
```

We should add tests for these cases and look for other places where this could be happening.

rafatower:
This is the classical one I'd tackle with TDD, but just my opinion.

jgoizueta:
Totally agree

rafatower:
After raising the quota the integration/regression tests look fine.

PR #329, opened 2016-12-20
Fix problem with period end dates

jgoizueta:
See #322


jgoizueta:
@ethervoid: your python wisdom would be appreciated here, and also an acceptance procedure

I couldn't figure out other places where we can have this problem, what do you think?

ethervoid:
Looks good to me

rafatower:
:+1:

Issue #324, opened 2016-12-15
Prevent type "geometry" does not exist when upgrading


rafatower:
When upgrading the extension we're getting some instances of `type "geometry" does not exist`. This happens because though the `geometry` type should be in the `search_path`, it is not.

My workaround for this was to edit the extension sql file and add this to the top:
```
SET search_path = "$user",public, cdb_dataservices_client;
```

Instead of that, it is best to add the schema explicitly and forget about this issue. (There's an issue related to this and moving extensions to different schemas but maybe not worth pursuing? I'll try to find and link here)

rafatower:
One of those errors:
```
[dbd-talos.useb.cartodb.net] out: 2016/12/15 12:26:58 [ERROR] [cartodb_user_d377e55d-4326-4faf-97c7-503535e81667_db pq: type "geometry" does not exist]
```

and investigating it a little:
```
cartodb_user_d377e55d-4326-4faf-97c7-503535e81667_db=# \dx
                                                   List of installed extensions
          Name           | Version |         Schema          |                             Description                             
-------------------------+---------+-------------------------+---------------------------------------------------------------------
 cartodb                 | 0.18.5  | cartodb                 | Turn a database into a cartodb user database.
 cdb_dataservices_client | 0.13.0  | cdb_dataservices_client | CartoDB dataservices client API extension
 crankshaft              | 0.5.1   | cdb_crankshaft          | CartoDB Spatial Analysis extension
 observatory             | dev     | cdb_observatory         | CartoDB Observatory backend extension
 odbc_fdw                | 0.2.0   | public                  | Foreign data wrapper for accessing remote databases using ODBC
 plpgsql                 | 1.0     | pg_catalog              | PL/pgSQL procedural language
 plproxy                 | 2.6.0   | observatory             | Database partitioning implemented as procedural language
 plpythonu               | 1.0     | pg_catalog              | PL/PythonU untrusted procedural language
 postgis                 | 2.2.2   | public                  | PostGIS geometry, geography, and raster spatial types and functions
 postgis_topology        | 2.2.2   | topology                | PostGIS topology spatial types and functions
 postgres_fdw            | 1.0     | public                  | foreign-data wrapper for remote PostgreSQL servers
(11 rows)
```
ok, postgis is there.

```
cartodb_user_d377e55d-4326-4faf-97c7-503535e81667_db=# \dT geometry
                     List of data types
 Schema |   Name   |               Description               
--------+----------+-----------------------------------------
[...]

PR #330, opened 2016-12-20
Set search path before installing/update the extension

jgoizueta:
See #324


jgoizueta:
@rafatower @ethervoid can you review?

rafatower:
:+1:

PR #334, opened 2016-12-29
Add cartodb to the search path

rafatower:
See
https://github.com/CartoDB/dataservices-api/issues/324#issuecomment-269614566

As the version `0.14.1` is not tagged and not released in production, I'm gonna try this. If the deploy in staging for testing does not work as expected, I'll generate a new version.

@jgoizueta can you please take a look?

### DIFF
--- a/client/cdb_dataservices_client--0.14.0--0.14.1.sql
+++ b/client/cdb_dataservices_client--0.14.0--0.14.1.sql
@@ -3,7 +3,7 @@
 \echo Use "ALTER EXTENSION cdb_dataservices_client UPDATE TO '0.14.1'" to load this file. \quit
 
 -- Make sure we have a sane search path to create/update the extension
-SET search_path = "$user",public,cdb_dataservices_client;
+SET search_path = "$user",cartodb,public,cdb_dataservices_client;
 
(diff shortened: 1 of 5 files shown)


client/cdb_dataservices_client--0.14.0--0.14.1.sql
```diff
@@ -3,7 +3,7 @@
 \echo Use "ALTER EXTENSION cdb_dataservices_client UPDATE TO '0.14.1'" to load this file. \quit
 
 -- Make sure we have a sane search path to create/update the extension
-SET search_path = "$user",public,cdb_dataservices_client;
+SET search_path = "$user",cartodb,public,cdb_dataservices_client;
```
jgoizueta: I would had placed `cartodb` after `public`, but I guess it'll work for our purposes any way.

rafatower: If you dig in the cartodb repo you'll find a long discussion about that with no conclusion. Should a user func in public take precedence over a cartodb func because of the search path? :)

The context of this change is just the lifespam of an extension installation/upgrade. So let me be lenient here with my own code and move on. I'll assume all the blame.

Issue #335, opened 2017-01-03
Broken quota unit tests, presumably because of 2017 NY bug


rafatower:
Both in master and development branches:
```
ubuntu@cdb:~/src/dataservices-api/server/lib/python/cartodb_services$ nosetests test/
...................................FF...............................................................................................
======================================================================
FAIL: I have 1000 credits and I just spent 1000 today. I should not pass
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/src/dataservices-api/server/lib/python/cartodb_services/test/metrics/test_quota.py", line 71, in test_routing_quota_check_fails_right_in_the_limit
    assert checker.check() == False
AssertionError

======================================================================
FAIL: test_routing_quota_check_fails_when_quota_exhausted (test.metrics.test_quota.TestQuotaChecker)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/src/dataservices-api/server/lib/python/cartodb_services/test/metrics/test_quota.py", line 53, in test_routing_quota_check_fails_when_quota_exhausted
    assert checker.check() == False
AssertionError

----------------------------------------------------------------------
Ran 132 tests in 0.319s

FAILED (failures=2)
```

rafatower:
There's a bug calculating the billing cycles in `__current_billing_cycle`:
```
> /home/ubuntu/src/dataservices-api/server/lib/python/cartodb_services/cartodb_services/metrics/user.py(81)__used_routing_quota()
-> date_from, date_to = self.__current_billing_cycle()
(Pdb) date_from
*** NameError: name 'date_from' is not defined
(Pdb) n
> /home/ubuntu/src/dataservices-api/server/lib/python/cartodb_services/cartodb_services/metrics/user.py(82)__used_routing_quota()
-> current_use = 0
(Pdb) date_from
datetime.date(2017, 1, 3)
(Pdb) date_to
datetime.date(2017, 1, 3)
(Pdb) 
```

rafatower:
In retrospective I think that `latest_valid_date` is a big smell when calculating billing cycles: https://github.com/CartoDB/dataservices-api/blob/ebbfae95cfca9414a5af3734ddab868221036eae/server/lib/python/cartodb_services/cartodb_services/metrics/user.py#L161

rafatower:
I tried a binary search of the problem but it was present from early versions of the code.

rafatower:
[...]

PR #337, opened 2017-01-04
Fix tests: zero-pad months in redis key #335

rafatower:


gfiorav:
The comment didn't post, don't know why. Here it is:

This will return `01` as string instead of previous `1` int (parsed to `1` string).

👍 if this is what you want

rafatower:
Yes, right what we want. Thanks!

PR #339, opened 2017-01-12
Obs_GetData OBS_GetMeta wrappers

talos:
This PR adds wrappers for the `OBS_GetData` and `OBS_GetMeta` Observatory functions that allow for much faster augmentation from the Observatory.

Because of issues with `RAISE NOTICE` causing a `CONTEXT` line in the test expectations but not in the setup I have, I can't validate tests yet.

This PR is not yet ready, but I've opened it open conversations around how this is done, as I haven't written these before.

### DIFF
--- a/client/renderer/interface.yaml
+++ b/client/renderer/interface.yaml
@@ -261,6 +261,38 @@
     - { name: boundary_id, type: text}
     - { name: time_span, type: text, default: 'NULL'}
 
+- name: obs_getdata
+  return_type: TABLE(id int, data json)
+  multi_row: true
+  multi_field: true
+  table_fields:
+    - { name: id, type: integer }
+    - { name: data, type: json }
+  params:
+    - { name: geomvals,  type: "geomval[]" }
+    - { name: params, type: json }
+    - { name: merge, type: boolean, default: true }
+
+- name: obs_getdata
+  return_type: TABLE(id text, data json)
+  multi_row: true
+  multi_field: true
+  table_fields:
+    - { name: id, type: text }
+    - { name: data, type: json }
+  params:
+    - { name: geomrefs,  type: "text[]" }
+    - { name: params, type: json }
+
+- name: obs_getmeta
+  return_type: json
+  params:
+    - { name: geom_ref,  type: "Geometry(Geometry, 4326)" }
+    - { name: params, type: json }
+    - { name: max_timespan_rank, type: integer, default: 'NULL' }
+    - { name: max_score_rank, type: integer, default: 'NULL' }
+    - { name: target_geoms, type: integer, default: 'NULL' }
+
 - name: obs_getcategory
   return_type: text
   params:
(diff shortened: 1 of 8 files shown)


talos:
This should now work, although I think Jenkins is failing because it builds against current version of observatory instead of 1.3.0 (https://github.com/CartoDB/observatory-extension/pull/246) which is what we need to use with these wrappers.

This is my first PR here, so very happy to spend some time walking through the process.

talos:
@rafatower We should look at this when we meet today.

ethervoid:
retest this please

client/renderer/interface.yaml
```diff
@@ -261,6 +261,38 @@
...
 
+- name: obs_getdata
+  return_type: TABLE(id int, data json)
+  multi_row: true
+  multi_field: true
+  table_fields:
+    - { name: id, type: integer }
+    - { name: data, type: json }
+  params:
+    - { name: geomvals,  type: "geomval[]" }
```
rafatower: Partially answering to myself: https://github.com/CartoDB/observatory-extension/pull/246/files/0f372604db40013a20ae0cea8ac14389cb7a0d2e#r98871206

talos:
Thank you for the review @rafatower !

> I'd like to know about perf tests with geomval arrays of different sizes (arrays in general). I've seen some screenshots but cannot find them.

@javisantana has the same curiosity.  My response is [here](https://github.com/CartoDB/camshaft/pull/254#issuecomment-276694894).  In short, our performance before this for complex geometries was already so problematic that I'm pretty sure we'll hit a query timeout before running out of space in a `geomval` array.  However, I can run some tests to explore this.

I don't believe there's any one-size fits all solution for this.  I've run tests where we send up one geometry at a time using `OBS_GetData` with a geomval array of length 1, and it's much slower.  There's a sweet spot whereby the table's geometries are chunked into geomval arrays, each of which is one `OBS_GetData` query.  However, that chunking needs to happen on the client side -- most easily, in Camshaft, and secondarily in documentation about how to use `OBS_GetData` for complex geometries.

> I have some concerns about geomval.

I addressed some concerns with `geomval` in [this comment](https://github.com/CartoDB/observatory-extension/pull/246/files/0f372604db40013a20ae0cea8ac14389cb7a0d2e#r98929261).  I'm fine with replacing our use of `geomval` with a custom type, I'm just not convinced of the value of doing so.

I can understand it's cleaner to have our own custom type, but is it worth the complexity of declaring that type in user accounts?  Keep in mind the custom type needs to exist at the client level, the server level, and the Observatory level.  Where should it be defined without causing a dependency hell?

Another possible approach might be to pass in anonymous `ROW` objects, although I'm not 100% sure of the feasibility (or cleanliness) of this solution.

> I have some concerns about how it should be charged.

This is a bug, I'd opened comments about this in Madrid but forgot to submit them in GitHub's PR review interface.  D'oh.  I'll fix those and add them to a PR where we can address these concerns, since this is merged.

> I'm missing some API docs in https://github.com/CartoDB/dataservices-api/tree/development/doc that also usually help reviewers as well as users

I don't think the documentation belongs there, it belongs in the observatory-extension.  You can find that documentation for these functions (in progress) [here](https://github.com/CartoDB/observatory-extension/blob/obs-getdata-getmeta-docs/doc/measures_functions.md#obs_getmetaextent-geometry-metadata-json).

rafatower:
> I don't think the documentation belongs there, it belongs in the observatory-extension. You can find that documentation for these functions (in progress) here.

This is the repo used to generate the public documentation sitting here:
https://carto.com/docs/carto-engine/dataservices-api/segmentation-functions

We'll have to create an issue ticket to deal with that.

talos:
Do you mean an issue to deal with moving the demographic functions currently documented in dataservices-api to observatory-extension, or more generally a ticket to make the observatory-extension docs live?

rafatower:
I mean a ticket to not duplicate API public documentation, regardless of where it lives. That involves other people.

Personally, I don't mind where the documentation lives as long as it is not duplicated.

talos:
Ah, I checked when I was writing before -- the documentation is not duplicated.  The snapshot functions are only described in dataservices-api, so we should be OK for now.

talos:
@ethervoid Has this been merged yet?  Attempting to run

```
CREATE TABLE my_complex_geometries AS
WITH meta AS (
  SELECT obs_getmeta(
    ST_MakeEnvelope(-179, 89, 179, -89, 4326),
    '[{"geom_id": "us.census.tiger.county"}]'::json
  ) AS meta)
SELECT row_number() over() AS cartodb_id, (data->0->>'value')::Geometry as the_geom
FROM obs_getdata(
  ARRAY[(ST_MakeEnvelope(-179, 89, 179, -89, 4326), 1)::geomval],
  (SELECT meta FROM meta), false);
```

yields `"function obs_getmeta(geometry, json) does not exist"`.

ethervoid:
It's merged in development, waiting for the comments of Rafa, but you can try it in staging cloud

rafatower:
@ethervoid what comments are you waiting for?

ethervoid:
None yet, I was talking in past. Now I've removed the quota usage part of the functions and pushed to development

Issue #340, opened 2017-01-25
Deploy major improvements to Observatory performance


talos:
### What we achieve

Deploy observatory-extension 1.3.0, which includes `OBS_GetData` and `OBS_GetMeta`, and then a new version of dataservices-api that wraps these two new functions.

* https://github.com/CartoDB/observatory-extension/pull/246/
* https://github.com/CartoDB/dataservices-api/pull/339

### Why

`OBS_GetData` and `OBS_GetMeta` provide a far more flexible and performant API for internal and external users to build off.  In particular:

* Considerable performance improvements for `OBS_GetMeasure` as used in the Builder UI
* Support for obtaining metadata for X columns and Y rows in 1 request, instead of X * Y requests
* Support for obtaining data for X columns and Y rows in 1 request, instead of X * Y requests

### Impact

* Immediate 20-30% speedup for existing OBS analysis
* After an update to Camshaft node, somewhere between a 3x and a 30x speedup for existing OBS interface
* Support for next-gen Observatory interface, which supports X columns at once (thus reducing strain on our servers & increasing usability immensely)

### Acceptance criteria

* Existing analysis in the UI continues to work
* Existing functions exposed via API for OBS still work as before (these are automatically tested)
* OBS_GetData and OBS_GetMeta work in user accounts

@ethervoid

talos:
(We should do this while I'm here)

ethervoid:
//cc @rafatower @rochoa

rafatower:
Added to the TODO as per our conversation. Yes, please, let's take advantage of working on the same timezone.

It is worth noting that the two tasks must go one after another in the order of the description.

@talos it'd be nice to cover in the issue description, very briefly, the following items: 
- what we try to achieve
- why
- impact
- acceptance criteria

so that anyone looking at the [Engine dashboard](https://github.com/orgs/CartoDB/projects/3#card-1489082) understands why we tackle it before other issues and hopefully other people on the team can contribute to get it to the Done column.

talos:
Yes, no problem.  Sorry for the brief description, I did not realize the purpose was to make the issue understandable more broadly -- now I understand.

### What we achieve

Deployment of observatory-extension 1.3.0, which includes `OBS_GetData` and `OBS_GetMeta`, and then a new version of dataservices-api that wraps these two new functions.

### Why

`OBS_GetData` and `OBS_GetMeta` provide a far more flexible and performant API for internal and external users to build off.  In particular:

[...]

PR #341, opened 2017-02-02
Release client 0.15.0 and server 0.21.0

ethervoid:
Related #340 

- Added functions `OBS_GetData` and `OBS_GetMeta`
- Added CONTRIBUTING document
- Isolines default provider changed to `mapzen` instead of `heremaps`


server/extension/sql/110_data_observatory_augmentation.sql
```diff
@@ -567,3 +567,151 @@ RETURNS NUMERIC AS $$
...
+  plpy.execute("SELECT cdb_dataservices_server._get_logger_config()")
+  logger_config = GD["logger_config"]
+  logger = Logger(logger_config)
+
+  with metrics('obs_getdata', user_obs_config, logger):
+    try:
+        obs_plan = plpy.prepare("SELECT * FROM cdb_dataservices_server._OBS_GetData($1, $2, $3, $4, $5);", ["text", "text", "geomval[]", "json", "boolean"])
+        result = plpy.execute(obs_plan, [username, orgname, geomvals, params, merge])
+        if result:
+          return result
```
talos: Shouldn't we be incrementing the OBS quota by the number of rows returned?  The only place where we should not be incrementing the quota is for calls to `obs_getmeta`.

I did some testing on this RE null rows, and our behavior is that we do not return the id/value set if we don't find data, so we don't have to worry about incrementing the success quota in cases where we couldn't find data for the client.

I did some testing and `len(result)` is usable, so code like:

    quota_service.increment_success_service_use(len(result))

should be OK.

server/extension/sql/110_data_observatory_augmentation.sql
```diff
@@ -567,3 +567,151 @@ RETURNS NUMERIC AS $$
...
+  plpy.execute("SELECT cdb_dataservices_server._get_logger_config()")
+  logger_config = GD["logger_config"]
+  logger = Logger(logger_config)
+
+  with metrics('obs_getdata', user_obs_config, logger):
+    try:
+        obs_plan = plpy.prepare("SELECT * FROM cdb_dataservices_server._OBS_GetData($1, $2, $3, $4);", ["text", "text", "text[]", "json"])
+        result = plpy.execute(obs_plan, [username, orgname, geomrefs, params])
+        if result:
+          return result
```
talos: Same comment as above, we should be incrementing quota by `len(result)`.

server/extension/sql/110_data_observatory_augmentation.sql
ethervoid: Ok, I'm a bit messy with this new functions.

So `obs_getmeta` doesn't need to increment the use of observatory quota, roger that.

The `obs_getdata` increment part, we should use the argument instead of the result because we should increment the same amount the success, empty and failed results. 

So, what do you think if we use the `len(geomvals)` as the number of results. If we do, we should compare the number of success results with the `geomvals`. If are the same all increment success results, if not `len(results)` goes for success and rest for empty.

What do you think?

ethervoid: One more thing is it possible to you to provide a call to the functions where a mix of empty and success responses are sent?

server/extension/sql/110_data_observatory_augmentation.sql
ethervoid: In this case instead of `geomvals` ,  should be `geomrefs` isnt't it?

server/extension/sql/110_data_observatory_augmentation.sql
ethervoid: You can see the change [here](https://github.com/CartoDB/dataservices-api/pull/341/commits/6337f84d9fd88a9958c92578b2f86c139da5086e)

server/extension/sql/110_data_observatory_augmentation.sql
ethervoid: You can see the change [here](https://github.com/CartoDB/dataservices-api/pull/341/commits/6337f84d9fd88a9958c92578b2f86c139da5086e)

server/extension/sql/110_data_observatory_augmentation.sql
```diff
@@ -603,19 +604,28 @@ RETURNS TABLE (
...
   with metrics('obs_getdata', user_obs_config, logger):
     try:
         obs_plan = plpy.prepare("SELECT * FROM cdb_dataservices_server._OBS_GetData($1, $2, $3, $4, $5);", ["text", "text", "geomval[]", "json", "boolean"])
         result = plpy.execute(obs_plan, [username, orgname, geomvals, params, merge])
+        empty_results = len(geomvals) - len(result)
         if result:
+          quota_service.increment_success_service_use(len(result))
           return result
         else:
+          quota_service.increment_empty_service_use(empty_results)
```
jgoizueta: Shouldn't we always `quota_service.increment_empty_service_use(empty_results)` ? (well, at least if `empty_results>0`).

I mean: we could have valid result and some empty results too, couldn't we?

server/extension/sql/110_data_observatory_augmentation.sql
```diff
@@ -651,21 +662,30 @@ RETURNS TABLE (
...
   with metrics('obs_getdata', user_obs_config, logger):
     try:
         obs_plan = plpy.prepare("SELECT * FROM cdb_dataservices_server._OBS_GetData($1, $2, $3, $4);", ["text", "text", "text[]", "json"])
         result = plpy.execute(obs_plan, [username, orgname, geomrefs, params])
+        empty_results = len(geomrefs) - len(result)
         if result:
+          quota_service.increment_success_service_use(len(result))
           return result
         else:
+          quota_service.increment_empty_service_use(empty_results)
```
jgoizueta: Same as above (make empty_results independently of result)

server/extension/sql/110_data_observatory_augmentation.sql
ethervoid: yep you're right. I was putting an easy example to check if that is what we want, but I'm fixing and testing it anyway :D

server/extension/sql/110_data_observatory_augmentation.sql
```diff
@@ -567,3 +567,173 @@ RETURNS NUMERIC AS $$
...
+  logger = Logger(logger_config)
+  quota_service = QuotaService(user_obs_config, redis_conn)
+  if not quota_service.check_user_quota():
+    raise Exception('You have reached the limit of your quota')
+
+  with metrics('obs_getdata', user_obs_config, logger):
+    try:
+        obs_plan = plpy.prepare("SELECT * FROM cdb_dataservices_server._OBS_GetData($1, $2, $3, $4);", ["text", "text", "text[]", "json"])
+        result = plpy.execute(obs_plan, [username, orgname, geomrefs, params])
+        empty_results = len(geomrefs) - len(result)
```
talos: In case unexpectedly the `len(result)` were greater than the `len(geomrefs)`, this could result in a negative `empty_results`.  We should make sure that a negative is replaced with `0`.

server/extension/sql/110_data_observatory_augmentation.sql
```diff
@@ -567,3 +567,173 @@ RETURNS NUMERIC AS $$
...
+  logger = Logger(logger_config)
+  quota_service = QuotaService(user_obs_config, redis_conn)
+  if not quota_service.check_user_quota():
+    raise Exception('You have reached the limit of your quota')
+
+  with metrics('obs_getdata', user_obs_config, logger):
+    try:
+        obs_plan = plpy.prepare("SELECT * FROM cdb_dataservices_server._OBS_GetData($1, $2, $3, $4, $5);", ["text", "text", "geomval[]", "json", "boolean"])
+        result = plpy.execute(obs_plan, [username, orgname, geomvals, params, merge])
+        empty_results = len(geomvals) - len(result)
```
talos: Same comment as comment above -- we need to handle possibility of a negative.

server/extension/sql/110_data_observatory_augmentation.sql
talos: To follow up, as @jgoizueta had mentioned this possibility, too -- `if empty_results:` is insufficient to protect against this, as that would evaluate to `true` for any nonzero number (including negative integers).

ethervoid: Ok, but it seems unlikely to have more results than geomvals.

ethervoid:
Anyway I wasn't able to get empty results and I love to have examples where empty results are mixed with success results in order to include in the acceptance tests

server/extension/sql/110_data_observatory_augmentation.sql
talos: But it will.  For example, though not documented yet, one use of this function will be to retrieve multiple geometries for a single geometry (the implementation of `OBS_GetBoundariesByGeometry` works through this.  Even if it didn't, we would want to make sure to catch such possible error cases, I think.

talos:
@ethervoid Hold on, I can get you an example of that.

server/extension/sql/110_data_observatory_augmentation.sql
ethervoid: Good to know, and that makes this check legit. Yes, it's good to catch this kind of problems but if they're going to happen. I don't like to make defensive programming

ethervoid:
Great, thank you :)

ethervoid:
@jgoizueta @talos Make a last review for this PR (you asked about some changes). If ok, we should move it to acceptance

talos:
@ethervoid Here's a query that will return fewer geometries than were passed in, due to input tracts not being able to join to gaps in the `us.census.tiger.cbsa` which is forced as the geometry:

```
with tracts as (
  select row_number() over () cartodb_id, *
  from cdb_dataservices_client.obs_getboundariesbygeometry(
    st_buffer(st_setsrid(st_point(-100, 44) , 4326), 2),
    'us.census.tiger.census_tract'
  )
), meta as (
  select cdb_dataservices_client.obs_getmeta(
    st_setsrid(st_extent(the_geom), 4326),
    '[{"numer_id": "us.census.acs.B01003001", "geom_id": "us.census.tiger.cbsa"}]',
    null, count(*)::INT
  ) meta
  from tracts
)
select * from cdb_dataservices_client.obs_getdata(
  (select array_agg((the_geom, cartodb_id)::geomval) from tracts),
  (select meta from meta)
);
```

I'm also going to provide an explicit example of a query that returns more geometries than were passed in (although as I said, `obs_getboundariesbygeometry` does that under the hood).

talos:
@ethervoid This query will return all tracts within a single CBSA, so there are more return rows than the length of the array passed in:

```
with cbsa as (
  select row_number() over () cartodb_id, *
  from cdb_dataservices_client.obs_getboundariesbygeometry(
    st_buffer(st_setsrid(st_point(-100, 44) , 4326), 2),
    'us.census.tiger.cbsa'
  )
  limit 1
), meta as (
  select cdb_dataservices_client.obs_getmeta(
    st_setsrid(st_extent(the_geom), 4326),
    '[{"geom_id": "us.census.tiger.census_tract"}]',
    null, count(*)::INT
  ) meta
  from cbsa
)
select * from cdb_dataservices_client.obs_getdata(
  (select array_agg((the_geom, cartodb_id)::geomval) from cbsa),
  (select meta from meta),
  false
);
```

talos:
@ethervoid Would love to see the examples above used in acceptance tests.

Besides that, the quotas look good to go now and I 👍 .

Issue #343, opened 2017-02-17
Make 3rd party services URL's configurable


rporres:
HERE has an offering of offline maps, so we need to be able to configure URL through `CDB_CONF`. It is currently hardcoded in https://github.com/CartoDB/dataservices-api/blob/development/server/lib/python/cartodb_services/cartodb_services/here/geocoder.py#L15 but prepared more or less to pass it as a parameter in https://github.com/CartoDB/dataservices-api/blob/development/server/lib/python/cartodb_services/cartodb_services/here/geocoder.py#L56

cc @javisantana 

**Update:** there are other services with hardcoded URL's.E.g: [mapzen isochrones]( https://github.com/CartoDB/dataservices-api/blob/7c816be5e5f99cd26d510774e1e23d1c4f282dfa/server/lib/python/cartodb_services/test/test_mapzenisochrones.py#L13). They should be configurable just like other settings (see the [configuration for different services](https://github.com/CartoDB/dataservices-api#local-install-instructions) ) and pick the hardcoded value as default only when absent.

javisantana:
good catch @rporres

javisantana:
hey, i'd like to see the approach here before moving on

rafatower:
We'll discuss general reqs for config stuff here in this ticket

jgoizueta:
# Background/Reference Information

We have three sources of dataservices configuration

* The Server (stored in DS server database in `cdb_conf` table)
* The User (stored in redis under `rails:users:{USER}` keys)
* The Organization (stored in redis under `rails:orgs:{ORG}` keys)

The DS server configuration is provisioned by chef, from node attributes assigned per environment:

* [production](https://github.com/CartoDB/cartodb-platform/blob/3af43aabfbeb280a39bdfd42c36fb6d5811cf773/environments/production.json#L195)
* [onpremises](https://github.com/CartoDB/cartodb-platform/blob/3af43aabfbeb280a39bdfd42c36fb6d5811cf773/environments/onpremises.json) (no ds config)
* [staging](https://github.com/CartoDB/cartodb-platform/blob/3af43aabfbeb280a39bdfd42c36fb6d5811cf773/environments/staging.json#L409)

The User and Organization configuration is managed by central, stored in its database and propagated through the boxes to redis metadata.

## Unfinished refactor

This refactor is in use for the Mapzen geocoding configuration.

It makes the configuration level (server, user, organization) and its storage independent (so configuration could be moved from db/redis etc.).
[...]

PR #347, opened 2017-03-02
Add configuration for Here and Mapzen services

jgoizueta:
See #343

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,15 +23,15 @@ Steps to deploy a new Data Services API version :
 
 ### Local install instructions
 
-- install data services geocoder extension 
+- install data services geocoder extension
 
     ```
     git clone https://github.com/CartoDB/data-services.git
     cd data-services/geocoder/extension
     sudo make install
     ```
-    
-- install observatory extension 
+
+- install observatory extension
 
     ```
     git clone https://github.com/CartoDB/observatory-extension.git
@@ -40,7 +40,7 @@ Steps to deploy a new Data Services API version :
     ```
 
 - install server and client extensions
-    
+
     ```
     # in data-services repo root path:
     cd client && sudo make install
@@ -51,7 +51,7 @@ Steps to deploy a new Data Services API version :
 
     ```
     # in data-services repo root path:
-    cd server/lib/python/cartodb_services && sudo pip install . --upgrade 
+    cd server/lib/python/cartodb_services && sudo pip install . --upgrade
     ```
 
 - install extensions in user database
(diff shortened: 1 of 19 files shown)


server/lib/python/cartodb_services/cartodb_services/here/routing.py
```diff
@@ -30,12 +30,17 @@ class HereMapsRoutingIsoline(Traceable):
         'quality'
     ]
 
-    def __init__(self, app_id, app_code, logger,
-                 base_url=PRODUCTION_ROUTING_BASE_URL):
+    def __init__(self, app_id, app_code, logger, service_params={}):
+        service_params = service_params or {}
```
rafatower: same as above here, regarding optional function args.

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -257,10 +265,21 @@ def heremaps_app_code(self):
...
+
+    @property
     def mapzen_matrix_api_key(self):
         return self._mapzen_matrix_api_key
 
     @property
+    def mapzen_matrix_service_params(self):
+        return self._mapzen_matrix_service_params
+
+    def mapzen_isochrones_service_params(self):
```
rafatower: for regularity's sake, shouldn't this be a `@property` as well?

server/lib/python/cartodb_services/test/test_heremapsgeocoder.py
```diff
@@ -145,3 +145,17 @@ def test_geocode_with_malformed_result(self, req_mock):
...
+        geocoder = HereMapsGeocoder(None, None, Mock(), { 'json_url': 'http://nonstandard_here_url' })
+        req_mock.register_uri('GET', 'http://nonstandard_here_url', text=self.GOOD_RESPONSE)
+        response = geocoder.geocode(
+            searchtext='Calle amor de dios',
+            city='Cordoba',
+            country='España')
+
+        self.assertEqual(response[0], -5.2794)
+        self.assertEqual(response[1], 37.70246)
+
```
rafatower: nice tests! :)

server/lib/python/cartodb_services/test/test_mapzengeocoder.py
```diff
@@ -109,3 +109,14 @@ def test_geocode_with_malformed_result(self, req_mock):
             self.geocoder.geocode(
                 searchtext='Calle Siempreviva 3, Valladolid',
                 country='ESP')
+
+    def test_geocode_address_with_nonstandard_url(self, req_mock):
+        nonstandard_url = 'http://nonstandardmapzen'
+        req_mock.register_uri('GET', nonstandard_url, text=self.GOOD_RESPONSE)
+        geocoder = MapzenGeocoder('search-XXXXXXX', Mock(), { 'base_url': nonstandard_url })
+        response = geocoder.geocode(
+            searchtext='Calle Siempreviva 3, Valldolid',
```
rafatower: minor issue: typo (somebody from Valladolid might be offended because of that misspelling :sweat_smile: )

server/lib/python/cartodb_services/cartodb_services/here/geocoder.py
```diff
@@ -52,14 +52,17 @@ class HereMapsGeocoder(Traceable):
         'strictlanguagemode'
         ] + ADDRESS_PARAMS
 
-    def __init__(self, app_id, app_code, logger, maxresults=DEFAULT_MAXRESULTS,
-                 gen=DEFAULT_GEN, host=PRODUCTION_GEOCODE_JSON_URL):
+    def __init__(self, app_id, app_code, logger, service_params={}, maxresults=DEFAULT_MAXRESULTS):
+        service_params = service_params or {}
```
jgoizueta: Yes, this looks redundant, but I added it because if the caller gets service_params from a dictionary (generated from the JSON stored in cdb_conf) and we could end up receiving a None here (if the `service` attribute doesn't exist).
Maybe we should make the default argument value `None` instead of `{}`to make the intent of that line more clear.

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
jgoizueta: Yes, it should be a @property, I missed it

server/lib/python/cartodb_services/test/test_mapzengeocoder.py
jgoizueta: Im my defence I have to say that I copy-pasted it from code written by a real vallisoletano 😁

server/lib/python/cartodb_services/cartodb_services/here/geocoder.py
rafatower: I see, I missed that case :+1:

jgoizueta:
@rafatower can you take another look? I took cae of the minor issues and added a bit of documentation.

README.md
```diff
@@ -86,6 +86,71 @@ Steps to deploy a new Data Services API version :
     SELECT CDB_Conf_SetConf('server_conf', '{"environment": "[development|staging|production]"}')
     ```
 
+- External services (Mapzen, Here) can have optional configuration, which is only needed for using non-standard services, such
```
rafatower: I'd take advantage of this to open a section `### Configuration` and another `### Optional configuration parameters` so that it is easier to read and to link to those sections.

Issue #346, opened 2017-03-01
Add ability to rate-limit geocodings per user


rafatower:
At present all users within an organization share the same pool of geocoding credits.

One particular user can simply exhaust all credits available in one session preventing other users of the same organization from performing any further geocodings until the billing period is finished.

We want to be able to limit geocoding calls per user and period of time.

In order to do so, a couple of configuration settings shall be added:
- `geocoder_user_rate_limit_limit`: number of requests allowed to the user per period of time
- `geocoder_user_rate_limit_period`: length of the period of time in seconds.

Those new settings shall be configurable at several levels:
- global (server configuration)
- organization
- user

User config shall take precedence over organization, and org config shall take precedence over global config. If it is absent, no limit should be applied (as it works at present).

E.g: the rate limit can be absent in global configuration but present at an organization level, with these values:
```python
{
  'geocoder_user_rate_limit_limit': 1000,
  'geocoder_user_rate_limit_period': 86400
}
```

That would mean that, regardless of the monthly quota allowance of the organization, no individual user within the organization could exceed 1000 geocodings/day.

**Edit 1:** instead of limiting "per day" make it generic to any configurable period. Added example as well.
**Edit 2:** this does not require UI. It is enough with a rake task + proper docs.

rafatower:
@javisantana a couple of questions:
- does this issue capture the requirements correctly?
- should we extend this to other services (I think so but maybe in another ticket)?

cc/ @rochoa and @jorgesancha for awareness.

javisantana:
I'd change

"We want to be able to limit geocoding calls per user and day."

by 

"We want to be able to limit geocoding calls per user and period of time"

So we configure the amount of time (in days or months)

jgoizueta:
Just some commentary about how to implement this.

I assume that to simplify the implementation we'll not try use a "rolling time window", but just use "static non-overlapping periods", i.e. partition time, so that we accumulate geocodings in time buckets, using floor(time_in_seconds / geocoder_user_rate_limit_period) as the time key.

I'll try to sound less pedantic 😛  : we won't count gecodings done during the last *geocoder_user_rate_limit_period* seconds, but just count geocodings in *geocoder_user_rate_limit_period*-second periods; when a new period starts the counter starts at 0.

So we can proceed as with the `UserConcurrentSyncsAmount` in amount where a counter is stored in a redis key `limits:Importer:UserConcurrentSyncsAmount:uUSERNAME`.

[...]

PR #355, opened 2017-03-14
Service rate limits

jgoizueta:
See #346 

Classes `ServiceManager` and `LegacyServiceManager` has been introduced to handle the configuration management and quota & rate limit checking. The Legacy counterpart class is intended for code not using the configuration refactor yet (which uses the GD configuration cache).

These classes also allow for unit-testing the configuration management (i.e. obtaining the server, user and organisation configuration for a given service/provider) and the rate limits.

This has been initially applied to the Mapzen and Here geocoding services.

Pending:

- [x] Apply to Google geocoding
- [x] SQL configuration interface

Nice to have:

- [ ] Apply to other services (thus cleaning up the code and enabling rate limits for them)
- [ ] Actual Rate-limit unit tests (*)
- [x] Refactor the rate-limits handling in legacy configuration (some clean-up/redundancy reduction could be achieved)

(*) Currently only the rate limit configuration is tested by unit tests. This is due to the rratelimit module we're using requiring Lua support to work with MockRedis. This is not critical since rratelimit has its own tests.


jgoizueta:
Notes (for reviewers)

I've introduced a `default` argument to the `get` methods of some (refactored) configuration objects, to allow specifying a default value like in the standard dictionary get method. This simplify its use in some cases.

But there's some inconsistency I've introduced in the default value for these arguments: in some cases the default `default` value is `KeyError`, which means raise a KeyError instead of returning a default value. I've done this to preserve existing behaviour (specially where cover by tests), but we'll need to review this.



 default=KeyError

server/extension/sql/20_geocode_street.sql
```diff
@@ -72,128 +72,96 @@ $$ LANGUAGE plpythonu;
...
-  # -- Check the quota
-  quota_service = QuotaService(user_geocoder_config, redis_conn)
-  if not quota_service.check_user_quota():
-    raise Exception('You have reached the limit of your quota')
+  service_manager = LegacyServiceManager('geocoder', username, orgname, GD)
+  service_manager.check()
 
   try:
-    geocoder = HereMapsGeocoder(user_geocoder_config.heremaps_app_id, user_geocoder_config.heremaps_app_code, logger, user_geocoder_config.heremaps_service_params)
+    geocoder = HereMapsGeocoder(service_manager.config.heremaps_app_id, service_manager.config.heremaps_app_code, service_manager.logger, service_manager.config.heremaps_service_params)
```
rafatower: having this encapsulated will let us decide on the caching of configs :+1:

server/extension/sql/20_geocode_street.sql
```diff
@@ -72,128 +72,96 @@ $$ LANGUAGE plpythonu;
...
+  service_manager = ServiceManager('geocoder', MapzenGeocoderConfigBuilder, username, orgname)
+  service_manager.check()
 
   try:
-    geocoder = MapzenGeocoder(mapzen_geocoder_config.mapzen_api_key, logger, mapzen_geocoder_config.service_params)
+    geocoder = MapzenGeocoder(service_manager.config.mapzen_api_key, service_manager.logger, service_manager.config.service_params)
     country_iso3 = None
     if country:
-      country_iso3 = country_to_iso3(country)
+      untry_iso3 = country_to_iso3(country)
```
rafatower: typo! (wof wof! :dog: I feel like houndci right know :D )

rafatower: I hope we have an integration test to catch that one

server/lib/python/cartodb_services/cartodb_services/refactor/storage/redis_config.py
```diff
@@ -9,11 +9,13 @@ def __init__(self, connection, config_key):
         self._config_key = config_key
         self._data = None
 
-    def get(self, key):
+    def get(self, key, default=KeyError):
```
rafatower: can we force all the config storages to have `default=None`?

server/lib/python/cartodb_services/cartodb_services/tools/exceptions.py
```diff
@@ -0,0 +1,3 @@
+class RateLimitExceeded(Exception):
```
rafatower: I'd rather define it near the point where it is raised (you know, high cohesion, loose coupling), but no strong opinion.

server/lib/python/cartodb_services/cartodb_services/tools/service_manager.py
```diff
@@ -0,0 +1,70 @@
...
+from cartodb_services.metrics import QuotaService
+from cartodb_services.tools import Logger
+from cartodb_services.tools import RateLimiter
+from cartodb_services.refactor.tools.logger import LoggerConfigBuilder
+from cartodb_services.refactor.core.environment import ServerEnvironmentBuilder
+from cartodb_services.refactor.backend.server_config import ServerConfigBackendFactory
+from cartodb_services.refactor.backend.user_config import UserConfigBackendFactory
+from cartodb_services.refactor.backend.org_config import OrgConfigBackendFactory
+from cartodb_services.refactor.backend.redis_metrics_connection import RedisMetricsConnectionFactory
+from cartodb_services.refactor.config.rate_limits import RateLimitsConfigBuilder
```
rafatower: In general I like to put all dependencies at the top of the file.

As they are dependencies of `ServiceManager`, that makes me think it deserves its own file.

Food for thought and a note for our future selves: one day the bottleneck will be object and class instantiations (not yet, no rush). Getting the dependencies right can be good for performance but also makes it easier to understand large codebases.

server/lib/python/cartodb_services/cartodb_services/tools/service_manager.py
```diff
@@ -0,0 +1,70 @@
...
+
+        self.rate_limiter = RateLimiter(rate_limit_config, redis_metrics_connection)
+        self.quota_service = QuotaService(self.config, redis_metrics_connection)
+
+from cartodb_services.metrics import QuotaService
+from cartodb_services.tools import Logger,LoggerConfig
+from cartodb_services.tools import RateLimiter
+from cartodb_services.refactor.config.rate_limits import RateLimitsConfigLegacyBuilder
+import plpy
+
```
rafatower: same as above here

server/lib/python/cartodb_services/cartodb_services/tools/service_manager.py
```diff
@@ -0,0 +1,70 @@
+from exceptions import RateLimitExceeded
+
```
rafatower: I'd start documenting responsibilities of new classes. In this case:
- pre-checks
- service config
- service quota management
- logger

server/lib/python/cartodb_services/cartodb_services/tools/service_manager.py
```diff
@@ -0,0 +1,70 @@
...
+class ServiceManagerBase:
+    def check(self, quota=True, rate=True):
```
rafatower: maybe a better name? I'd use `assert_whatever` rather than `check` cause it usually conveys the idea that if the check fails it'll raise an exception or somehow stop there (instead of delegating that on the caller).

Proposal: `assert_within_limits`

my 2 cents

server/lib/python/cartodb_services/cartodb_services/refactor/storage/redis_config.py
jgoizueta: That would change the default behaviour so could introduce unknown problems (it actually breaks some tests). So I'll keep this as is to avoid taking care of it at the moment. It would be nice to review this at a later time, though, making the default None as in other places.

server/extension/sql/20_geocode_street.sql
jgoizueta: I hope too that would be catch by the integration tests, the other tests don't check it. I ran some manual local tests for the SQL interface but I probably introduced that error after testing 😅

server/lib/python/cartodb_services/cartodb_services/refactor/storage/redis_config.py
rafatower: :+1:

jgoizueta:
Should we add a `rate_limits` column to  `cdb_service_quota_info`? (with the rate limits configuration for each service)

@rafatower @ethervoid what do you think?

rafatower:
Let's keep quota and rate limits separated as far as we can.

jgoizueta:
Notes for reviewers:

There's some pretty extensive changes to the template renderer (`SqlTemplateRenderer`) and client functions templates. This where originated in adding support for `void` return values and client private functions (with username, orgname parametrs). On the bright side, the templates are much simpler now (hopefully not less readable), and I also managed to avoid changing the tests output 😅 

We could do without the client configuration functions, but I think they can make configuration easier in some cases (e.g. through a script or rake task).

client/renderer/templates/30_plproxy_functions.erb
```diff
@@ -1,9 +1,9 @@
-CREATE OR REPLACE FUNCTION <%= DATASERVICES_CLIENT_SCHEMA %>._<%= name %> (<%= ['username text', 'organization_name text'].concat(params_with_type_and_default).join(', ') %>)
+CREATE OR REPLACE FUNCTION <%= DATASERVICES_CLIENT_SCHEMA %>._<%= name %> (<%= params_with_type_and_default(true).join(', ') %>)
```
rafatower: I don't quite understand why we're passing `true` in this case.

doc/rate_limits.md
```diff
@@ -0,0 +1,110 @@
...
+(i.e. if more than `limit` calls are performe in a fixed interval of
+duration `period` seconds) the call will fail with an "Rate limit exceeded" error.
+
+## Server-side interface
+
+There's a server-side SQL interface to query or change the configuration.
+
+### cdb_dataservices_server.cdb_service_get_rate_limit(username, orgname, service)
+
+This function returns the rate limit configuration for a given user and service.
```
rafatower: does it return..:
- (a) the configuration at the user level if it exists?
- (b) or the *effective configuration* applied to that user? (that is, the org or the server rate limit if there's no user rate limit defined)

I'd say we need both with clear names to avoid confusion.

doc/rate_limits.md
```diff
@@ -0,0 +1,110 @@
...
+
+### cdb_dataservices_client.cdb_service_set_server_rate_limit(username, orgname, service, rate_limit)
+
+This function sets the default rate limit configuration for all users accesing the dataservices server. This is overriden by organization of user configuration.
+
+The configuration is provided as a JSON literal. To remove the organization-level configuration `NULL` should be passed as the `rate_limit`.
+
+#### Returns
+
+This functions doesn't return any value.
```
rafatower: Pretty neat. If anything other than the comment above, I'd miss some example

server/extension/sql/20_geocode_street.sql
```diff
@@ -72,128 +72,96 @@ $$ LANGUAGE plpythonu;
...
+  service_manager = ServiceManager('geocoder', MapzenGeocoderConfigBuilder, username, orgname)
+  service_manager.assert_within_limits()
 
   try:
-    geocoder = MapzenGeocoder(mapzen_geocoder_config.mapzen_api_key, logger, mapzen_geocoder_config.service_params)
+    geocoder = MapzenGeocoder(service_manager.config.mapzen_api_key, service_manager.logger, service_manager.config.service_params)
     country_iso3 = None
     if country:
-      country_iso3 = country_to_iso3(country)
+        country_iso3 = country_to_iso3(country)
```
rafatower: uhm, are you consistent with spaces? wouldn't mention if it weren't python

server/extension/sql/210_rates.sql
```diff
@@ -0,0 +1,91 @@
...
+  service TEXT)
+RETURNS JSON AS $$
+  import json
+  from cartodb_services.config import ServiceConfiguration, RateLimitsConfigBuilder
+
+  import cartodb_services
+  cartodb_services.init(plpy, GD)
+
+  service_config = ServiceConfiguration(service, username, orgname)
+  rate_limit_config = RateLimitsConfigBuilder(service_config.server, service_config.user, service_config.org, service=service, user=username, org=orgname).get()
```
rafatower: note to myself: is this the effective rate limit config?

rafatower: ok, it's the effective configuration

server/lib/python/cartodb_services/cartodb_services/config/rate_limits.py
```diff
@@ -0,0 +1,114 @@
...
+    def is_limited(self):
+        return self._limit and self._limit > 0 and self._period and self._period > 0
+
+class RateLimitsConfigBuilder(object):
+    """
+    Build a rate limits configuration obtaining the parameters
+    from the user/org/server configuration.
+    """
+
+    # TODO: user->username, org->orgname
```
rafatower: Time to change it?

client/renderer/sql-template-renderer
```diff
@@ -17,7 +17,7 @@ class SqlTemplateRenderer
   end
 
   def render
-    ERB.new(@template).result(binding)
+    ERB.new(@template, nil, '-').result(binding)
```
jgoizueta: Good practice, I'm all for it. And looking at this case now I realise that I used that trim_mode while developing but ended up not needing it, so I think I'll better remove it.

client/renderer/sql-template-renderer
```diff
@@ -44,16 +44,29 @@ class SqlTemplateRenderer
...
+    parameters = []
+    if with_credentials
+      parameters << { 'name' => 'username', 'type' => 'text' }
+      parameters << { 'name' => 'orgname',  'type' => 'text' }
+    end
+    parameters + @function_signature['params'].reject(&:empty?)
+  end
+
+  def credentials_declaration()
+    "username text;\n  orgname text;" if public_function?
```
jgoizueta: Agreed, let's avoid confusion, it should be `user_org`, not `credientials`.

client/renderer/sql-template-renderer
```diff
@@ -62,6 +75,49 @@ class SqlTemplateRenderer
     end
     return parameters
   end
+
+  def public_function?
+    !@function_signature['private']
```
jgoizueta: `public` here means it is granted to the public role, so can be used by regular user roles. I left it as the default since all template-generated function until now where of this kind.

`private` means accessible only to the super user, `postgres`. So these functions need explicit user/org arguments to identify the user. This is used for the new rate limit configuration changing functions.

client/renderer/sql-template-renderer
```diff
@@ -62,6 +75,49 @@ class SqlTemplateRenderer
...
+    return_type.downcase == 'void'
+  end
+
+  def return_declaration
+     "ret #{return_type};" unless void_return_type? || multi_row
+  end
+
+  def return_statement(&block)
+    if block
+      erbout = block.binding.eval('_erbout')
```
jgoizueta: :+1:

client/renderer/templates/30_plproxy_functions.erb
jgoizueta: I should add a parameter name to make this clear. The original parameter name is `with_credentials` which is a misnomer as discussed above, so I'll change it.

As we are generating here the proxies for the server functions, all these functions have the user/org parameters.

client/renderer/sql-template-renderer
rafatower: I see. Since we did not have any "superuser" function before and taking security into account, what do you think is the best default?

PS: maybe `superuser` is a better name/qualifier for those functions.

doc/rate_limits.md
jgoizueta: It provides the second (b) effective configuration. I thought we could do without the individual server/org/user configuration information. Let's discuss this in a few minutes.

client/renderer/templates/30_plproxy_functions.erb
rafatower: got it, thanks for the clarification :+1:

server/extension/sql/20_geocode_street.sql
jgoizueta: Mmm seems (valid but) incosistent with the rest 2-char indentation. I had some troubles with my editor's config 😅

doc/rate_limits.md
jgoizueta: As per our conversation, I'm keeping just the single function that retrieves the *effective* configuration and I've described this behaviour in the documentation.

client/renderer/sql-template-renderer
jgoizueta: As we discussed, I'm changing the terminology to `superuser` functions (rather than private) which is more descriptive of the actual differenciation, and we'll keep the default to be non-superuser (public) which is what all previously existing template functions were.

jgoizueta:
Hey @rafatower can you take a final look? If everything's OK I'll then merge into development to start with acceptance checks (in staging).

PR #360, opened 2017-04-10
Add geom_tags to getavailablegeometries

talos:
Also, adds a few additional columns to obs_getavailabletimespans.

### DIFF
--- a/server/extension/sql/130_data_observatory_helper.sql
+++ b/server/extension/sql/130_data_observatory_helper.sql
@@ -35,7 +35,7 @@ RETURNS SETOF cdb_dataservices_server.obs_meta_denominator AS $$
   SELECT * FROM cdb_observatory.OBS_GetAvailableDenominators(bounds, filter_tags, numer_id, geom_id, timespan);
 $$ LANGUAGE plproxy;
 
-CREATE TYPE cdb_dataservices_server.obs_meta_geometry AS (geom_id text, geom_name text, geom_description text, geom_weight text, geom_aggregate text, geom_license text, geom_source text, valid_numer boolean, valid_denom boolean, valid_timespan boolean, score numeric, numtiles bigint, notnull_percent numeric, numgeoms numeric, percentfill numeric, estnumgeoms numeric, meanmediansize numeric);
+CREATE TYPE cdb_dataservices_server.obs_meta_geometry AS (geom_id text, geom_name text, geom_description text, geom_weight text, geom_aggregate text, geom_license text, geom_source text, geom_type text, geom_extra jsonb, geom_tags jsonb, valid_numer boolean, valid_denom boolean, valid_timespan boolean, score numeric, numtiles bigint, notnull_percent numeric, numgeoms numeric, percentfill numeric, estnumgeoms numeric, meanmediansize numeric);
 
 CREATE OR REPLACE FUNCTION cdb_dataservices_server.OBS_GetAvailableGeometries(
   username TEXT,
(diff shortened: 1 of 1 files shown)


talos:
This should be deployed at the same time as https://github.com/CartoDB/observatory-extension/pull/282/files .  @rochoa let me know when this is reviewed so we can coordinate that.

talos:
This should be deployed at the same time as https://github.com/CartoDB/observatory-extension/pull/282/files .  @rochoa let me know when this is reviewed so we can coordinate that.

rochoa:
Assigning the review to @rafatower. I'm not sure if we can change a type in that way.

rochoa:
@rafatower, thanks a lot for the detailed, and quick, CR.

talos:
I can write the upgrade/downgrade paths, but disagree that we should generate new API functions.  I think even a few hours of downtime are acceptable, because:

1. Our UI in its current form does not depend on these functions (it depends on `OBS_GetAvailableNumerators`, so breaking them briefly will not cause any downtime there.
2. No publicly available API functions except these two (`OBS_GetAvailableGeometries` and `OBS_GetAvailableTimespans`) will be affected during the downtime.
3. While they are documented and usable, actual usage of `OBS_GetAvailableGeometries` and `OBS_GetAvailableTimespans` at the moment is factually only me, as evidenced by kibana.

However, if we instead generate entirely new APIs to avoid downtime for functions that nobody is currently using, we will end up with:

1. A bunch of orphan, inconsistent, and useless API functions (these two)
2. A UI that depends on very slightly altered and named versions of those functions, which are then in fact consistent with `OBS_GetAvailableNumerators` and `OBS_GetAvailableDenominators`.

It's also important to note that our current documentation of `OBS_GetAvailableGeometries` actually [claims that we return a geom_tags column](https://carto.com/docs/carto-engine/data/discovery-functions#obsgetavailablegeometriesbounds-filtertags-numerid-denomid-timespan), so this PR is for that function simply *bringing functionality in line with what we claim we already do*.

I recognize that changing the signatures of these functions is a sin as these are publicly accessible APIs.  However, I think it is still the right choice, especially because they will _actually_ be put into use soon for the new UI.

@rochoa @rafatower thoughts?

rafatower:
Taking into account that:
- `OBS_GetAvailableGeometries` and `OBS_GetAvailableTimespans` are the only affected functions
- the usage data you provide

and considering this an exception, I'd just assume that those functions will be broken for a while.

Then, I would:
1. create an issue (a bug) in this repo to track and go through the whole process (including review, acceptance and deployment). Let us deal with upgrade/downgrade and the deployment of both server and client.
2. Deploy the changes in OBS
3. Deploy the changes in DS server and DS client

How does it sound?

talos:
Sounds good to me, @rafatower .  I'll create the bug today.  Want me to tag you in it?

Once the changes are deployed in OBS, I'll make a note in that bug.

rafatower:
OK. Better tag rochoa instead for prioritization. Thank you.

rafatower:
Tests are no problem in this instance. See https://github.com/CartoDB/dataservices-api/issues/364

talos:
Great.  Please ping me when this is deployed.

talos:
What's the status of this?  Calling `select * from obs_getavailablegeometries(st_setsrid(st_point(-73.9,40.7), 4326))` over SQL API right now is still returning the error:

```
cdb_dataservices_client._obs_getavailablegeometries(7): [cartodb_user_a6f0d0fe-4f4f-4217-8150-22b0010fe409_db] REMOTE ERROR: PL/Proxy function cdb_dataservices_server.obs_getavailablegeometries(7): Got too many fields from remote end
```

@rafatower

PR #362, opened 2017-04-12
Fix the rate limits documentation

jgoizueta:
The service name used to store the geocoding rate limits is geocoder, not geocoding.


jgoizueta:
@rafatower FYI: The documentation of rate limits had the service name wrong, because we didn't have enough confusion about such things yet.

I'm fixing this directly in master then I'll merge into development.

rafatower:
ok

Issue #371, opened 2017-05-25
Dataservices geocoder fallback issues for namedplaces


jvillarf:
## Description
**Product**: onpremises-2.1.0
**Related ticket**: https://github.com/CartoDB/onpremises/issues/372

@ethervoid has well explained it here: https://github.com/CartoDB/onpremises/issues/372#issuecomment-304015260

Basically, we are **trying** to geocode **namedplaces** using **mapzen**, even if it hasn't been configured as a potential external geocoder, causing the following error in **builder**:
````
curl -k "https://carto.lan/user/carto/api/v2/sql?q=select%20*%20from%20cdb_dataservices_client.cdb_service_quota_info()%3B&api_key=712cffb5f528b44df0bd3a740d0be555ac5531aa"
{"error":["cdb_dataservices_client._cdb_service_quota_info(2): [db_dataservices] REMOTE ERROR: spiexceptions.ExternalRoutineException: cartodb_services.metrics.config.ConfigException: Mapzen config is not setted up"]
````

As a result, **builder** is not showing the proper options inside **analysis**.

![smoke1](https://cloud.githubusercontent.com/assets/15977491/26453752/15f09408-4165-11e7-8ddb-0a2cb6c329c6.png)

Having an **onpremises-2.1.0** with **LDS support** just for internal geocoding/Data observatory will lead to this issue, unless we set a **fake mapzen geocoder API key** at installation time.

@juanignaciosl @rafatower  Unless a really big refactor is needed to fix it, i'll consider it as blocking for the **Onpremises-2.1.0** release.

Thanks.

jvillarf:
## CR and review

I've just been playing around with **onpremises** ```dataservices-1.0.0-pre3``` that includes this very same fix and it's working as expected:

1- **No mapzen API key**: named places use **internal geocoder** and there is no sign of **mapzen** in the logs.
```` bash
[...]

PR #373, opened 2017-05-26
Fix for namedplaces geocoding functions fallback

ethervoid:
Related #371 

Also I've added a check for the google client credentials to avoid an ugly error message.


server/extension/sql/50_namedplaces.sql
```diff
@@ -1,21 +1,23 @@
 ---- cdb_geocode_namedplace_point(city_name text)
 CREATE OR REPLACE FUNCTION cdb_dataservices_server.cdb_geocode_namedplace_point(username text, orgname text, city_name text)
 RETURNS Geometry AS $$
+  import spiexceptions
   try:
     mapzen_plan = plpy.prepare("SELECT cdb_dataservices_server._cdb_mapzen_geocode_namedplace($1, $2, $3) as point;", ["text", "text", "text"])
     return plpy.execute(mapzen_plan, [username, orgname, city_name])[0]['point']
-  except BaseException as e:
+  except spiexceptions.ExternalRoutineException as e:
```
rafatower: I don't understand how this PR fixes the mapzen issue.

The point of the issue is to not try mapzen at all if it is not configured correctly.

rafatower: ok, as per @ethervoid the exception is raised here: https://github.com/CartoDB/dataservices-api/blob/ddcc7162fb54f6276b655bb2f6c7ced863aa9f40/server/lib/python/cartodb_services/cartodb_services/metrics/config.py#L368

but it is wrapped in an `spiexceptions.ExternalRoutineException` because of the call through `plpy` module.

There will be a ticket for a better config fix (if there isn't one yet).

:+1:

rafatower:
Fine by me after fixing the tests. Thanks a lot for the quick response

server/extension/sql/50_namedplaces.sql
ethervoid: Here it is #374

PR #376, opened 2017-05-31
Changes for local installation

juanignaciosl:
This adds needed changes at the configuration.

I  addition, I'm facing some issues while I try to load the geocoder dump.

If I follow the instructions, I run `geocoder_restore_dump` at `dataservices_db` and it fails:

```
psql:db_dumps/ne_admin0_v3.sql:293: ERROR:  Missing "public"._CDB_UserQuotaInBytes()
```

I thought that it might be caused by an inconsistency on the documentation, so I followed [geocoder installation instructions](https://github.com/CartoDB/data-services/tree/master/geocoder/extension) in order to run the dump into a CARTO user account instead of `dataservices_db`. Anyway, I got another error:

```
psql:db_dumps/ne_admin0_v3.sql:293: ERROR:  function _cdb_nonanalysistablesinschema(text) does not exist
LINE 6:     SELECT table_name FROM _CDB_NonAnalysisTablesInSchema(sc...
                                   ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

What am I doing wrong? cc @rafatower @ethervoid

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,14 @@ Steps to deploy a new Data Services API version :
      psql -U postgres -d dataservices_db -c "BEGIN;GRANT SELECT ON ALL TABLES IN SCHEMA observatory TO dataservices_user; COMMIT" -e
      psql -U postgres -d dataservices_db -c "BEGIN;GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA observatory TO dataservices_user; COMMIT" -e
      ```
+     
+   - Configure geocoder API at `config/app_config.yml`:
+     ```
+     host: 'localhost'
+     port: '5432'
+     user: 'dataservices_user'
+     dbname: 'dataservices_db'
+     ```
 
 ### Server configuration
 


rafatower:
The geocoder API is independent from Builder. So IMHO that doc belongs in CartoDB repo, as it is now IIRC. Adding it here is a bit redundant and out of scope.

juanignaciosl:
> You shall install the internal geocoder on top of a fully-fledged CARTO account, as stated in docs (...)

Yes, I know. The problem is that following [Dataservices API instructions section at "install internal geocoder"](https://github.com/CartoDB/dataservices-api/blob/8ea1ffc42510d0b0499c1097fec8ec05bd9989c5/README.md) in order won't work, because geocoder installation is mentioned later, and because it refers to `dataservices_db` as the database for the geocoder dump instead of that user database.

ethervoid:
Why don't we remove [this triggers](https://github.com/CartoDB/data-services/blob/master/geocoder/extension/sql/0.0.1/20_admin0.sql#L182-L183) taking into account that we're not going to check quota for that table...Neither internal geocoder nor dataservices-api should require a fully-fledged CARTO account.

rafatower:
DS API offers an API to different services. Builder is "just one possible client". The internal geocoder is "just one of the many services". I'd like to keep the conceptual separation, loose coupling and segregation of documentations as much as possible.

Removing the dependency upon CARTO account is "technical debt" and should have its own ticket to go through the whole process (including acceptance), you know... I added a ticket for it (and will include it in the *what's next* as well): https://github.com/CartoDB/data-services/issues/229

ethervoid:
Yes, I agree with both points. I was just pointing something that could be made, removing the carto dependency from data-services, and if it was a feasible change then create the ticket as you did :)

PR #377, opened 2017-05-31
Removed wrong square brackets

juanignaciosl:
👀  @ethervoid @rafatower

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,8 @@ Steps to deploy a new Data Services API version :
 
 Configuration for the different services must be stored in the server database using `CDB_Conf_SetConf()`.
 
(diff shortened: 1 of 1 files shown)


rafatower:
`[square brackets]` means optional in this context. Almost sure it is indeed optional.
Is the enclosed config mandatory or optional, @ethervoid ?
Any other suggested syntax if they were removed, @juanignaciosl ?

ethervoid:
Yes, that config is inside square brackets because its optional

juanignaciosl:
Oh. I was confused because it was too close to valid JSON and there are no other optional parameters. Maybe mentioning in plain text previously is enough. Thanks!

PR #379, opened 2017-05-31
Missing GRANT USAGE

juanignaciosl:
👀  @rafatower @ethervoid


rafatower:
Shall we not change it to the actual `geocoder_api` user, which is granted usage upon installation of the extension? or shall we not change that user by a `dataservices_api` user instead?

juanignaciosl:
I don't know the best answer, but I can be useful as a rubber duck :_)

How come `geocoder_api`? It's not even mentioned in this documentation.

What would be the difference between using `dataservices_api`  user and `dataservices_user`?

rafatower:
It *is* `geocoder_api` for historical reasons: https://github.com/CartoDB/dataservices-api/blob/development/server/extension/sql/999_geocoder_server_user.sql

That was one of my points: we're kind of "fixing" something by documenting something else that diverges from the intended design and code, names aside.

My other point is that maybe we shall change that name for clarity. Names matter.

juanignaciosl:
Oh, +1 to go back to `geocoder_api`.

Although names do matter, doing a rename now might make everything more complicated, maybe this is not the moment for a naming improvement.

ethervoid:
Ok, lets move all the `dataservices_user` to `geocoder_api`

PR #390, opened 2017-07-21
Fix typo while using logger

dgaubert:


dgaubert:
❤️  @ethervoid

juanignaciosl:
As this has been around for 9 months I think that it can wait for the next dataservices release.

PR #392, opened 2017-08-08
Precheck functions for DO

ethervoid:
New functions to perform a pre-check for the user data and metadata
parameters that are going to be used to execute an OBS_GetData operation.

Doing this before the OBD_GetData we are able to check if we're going to
have problems with the current request. Is useful because right now
we're using the *_Exception_Safe functions which return NULL in case we
have an exception, this leads to execute the whole request even if we
know beforehand that it's going to fail.

The _OBS_PreCheck function is private because I don't want to expose it as public yet.

I have to include the upgrade/downgrade code but I need a CR for verification


client/sql/35_precheck_functions.sql
```diff
@@ -0,0 +1,60 @@
...
+        END IF;
+
+        IF geoms.valid IS FALSE THEN
+            errors := array_append(errors, 'There are invalid geometries in the input data, please review them');
+        END IF;
+
+        -- 1000 vertex for a geometry is a limit we have in the obs_getdata function. You can check here
+        -- https://github.com/CartoDB/observatory-extension/blob/1.6.0/src/pg/sql/41_observatory_augmentation.sql#L813
+        IF geoms.avg_vertex > 1000 THEN
+            errors := array_append(errors, 'The average number of geometries vertex is greater than 1000, please try to simplify them');
```
jgoizueta: Alternative wording suggestion: "The average number of vertices per geometry is greater than 1000,..."

Bonus pedant-nerd info: vertices is the plural of vertex, but if you prefer vertexes, that's alright too!

ethervoid: Like it, changing!

PR #394, opened 2017-08-11
Updated configuration

antoniocarlon:
Updated the User database configuration section

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Steps to deploy a new Data Services API version :
      ```
    - Give permission to execute and select to the `dataservices_user` user:
      ```
-     psql -U postgres -d dataservices_db -c "BEGIN;CREATE EXTENSION IF NOT EXISTS observatory version 'dev'; COMMIT" -e
+     psql -U postgres -d dataservices_db -c "BEGIN;CREATE EXTENSION IF NOT EXISTS observatory VERSION 'dev'; COMMIT" -e
      psql -U postgres -d dataservices_db -c "BEGIN;GRANT SELECT ON ALL TABLES IN SCHEMA cdb_observatory TO dataservices_user; COMMIT" -e
      psql -U postgres -d dataservices_db -c "BEGIN;GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA cdb_observatory TO dataservices_user; COMMIT" -e
      psql -U postgres -d dataservices_db -c "BEGIN;GRANT SELECT ON ALL TABLES IN SCHEMA observatory TO dataservices_user; COMMIT" -e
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,7 +92,7 @@ Steps to deploy a new Data Services API version :
      ```
    - Give permission to execute and select to the `dataservices_user` user:
      ```
-     psql -U postgres -d dataservices_db -c "BEGIN;CREATE EXTENSION IF NOT EXISTS observatory; COMMIT" -e
+     psql -U postgres -d dataservices_db -c "BEGIN;CREATE EXTENSION IF NOT EXISTS observatory version 'dev'; COMMIT" -e
```
javitonino: Nitpicky, but can you make `version` uppercase? For consistency with the rest of SQL queries.

README.md
```diff
@@ -254,14 +254,16 @@ jsonb_set(
 ```
 ### User database configuration
 
+#### Option 1
```
javitonino: Maybe `#### Option 1 (manually)`

README.md
```diff
@@ -272,10 +274,31 @@ SELECT CDB_Conf_SetConf(
...
-#### Search path
+##### Search path
 
 The search path must be configured in order to be able to execute the functions without using the schema:
 
 ```sql
 ALTER ROLE "<USER_ROLE>" SET search_path="$user", public, cartodb, cdb_dataservices_client;
 ```
+
+#### Option 2
```
javitonino: Maybe `#### Option 2 (from Builder)`

README.md
```diff
@@ -272,10 +274,31 @@ SELECT CDB_Conf_SetConf(
...
+
+In the `cartodb/config/app_config.yml` file, enable the desired dataservices:
+```
+enabled:
+    geocoder_internal: false
+    hires_geocoder: false
+    isolines: false
+    routing: false
+    data_observatory: true
+```
```
javitonino: I think we are missing a rake execution, aren't we?

javitonino:
@ethervoid Mind taking a final look as out local dataservices expert? 🙏

ethervoid:
I don't think the option 2 should be included in the public readme. In that case, we have to include in the installation steps but we shouldn't mix the DS readme (which shouldn't know about other projects) with the installation documentation

javitonino:
Well, we need to document this somehow. We could add it to the installation guide http://cartodb.readthedocs.io/en/latest/operations/configure_data_services.html if that sounds better to you. I'd still think it'd be nice to to link it from here though.

We can pretend that DS API is independent from CARTO, but not really. But I agree that separating it to the installation guide is good, this README is pretty long already.

PR #399, opened 2017-09-07
Keep metrics keys consistent

rafatower:
See cartodb repo and our internal documentation:
https://github.com/CartoDB/cartodb/search?utf8=%E2%9C%93&q=failed_responses&type=


jgoizueta:
How are we going to handle the variant keys when this is deployed?

rafatower:
At present, to get the total you need to sum both `fail_responses` (from DS API) and `failed_responses` (from cartodb queue servers). This won't change in the short term. Eventually we'd just need to read `failed_responses`, which is what's in the documentation.

PR #400, opened 2017-10-04
Upgrade googlemaps python library to v2.5.1

### LINKED ISSUE #383: Upgrade google services client library to v2.5.1
While investigating an issue related to request signing and not being accepted by google services, we figured out that for some reason current google services client library was not producing valid signatures for requests issued from DS API when the secret key contains a dash character.

See this issue: https://github.com/CartoDB/dataservices-api/issues/382

Upgrading the `googlemaps` library fixed the problem.

The dependency  on the particular version of that library is coded here:
https://github.com/CartoDB/dataservices-api/blob/402c14e6cabc1c9512d1fa59ab78fd12f0d1be63/server/lib/python/cartodb_services/requirements.txt#L4

cc/ @ernesmb

rafatower:
This closes #383

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -21,7 +21,7 @@ DATA = $(NEW_EXTENSION_ARTIFACT) \
 SOURCES_DATA_DIR = sql/
 
 
-REGRESS = $(notdir $(basename $(wildcard test/sql/*test.sql)))
+REGRESS = $(notdir $(basename $(sort $(wildcard test/sql/*test.sql))))
 TEST_DIR = test/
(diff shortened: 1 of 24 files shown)


rafatower:
The first commit is upgrading the dependency. The rest is just making the tests compatible with PG9.5 and 9.6 and make them more robust.

cc/ @juanignaciosl @javitonino @ethervoid

javitonino:
Oh, that's probably what is happening with `data-services` (geocoder) tests as well. I'll see if I can copy some of your changes over there to make tests run. Thanks!

Issue #382, opened 2017-06-15
Fix google services secret key validity check


rafatower:
It turns out that the google keys are not strictly base64-encoded but a variant of it:

https://developers.google.com/maps/documentation/directions/get-api-key

> 3. Retrieve your private key, which is encoded in a modified Base64 for URLs, and sign the URL above using the HMAC-SHA1 algorithm. You may need to decode this key into its original binary format. Note that in most cryptographic libraries, the resulting signature will be in binary format.
>
> Note: Modified Base64 for URLs replaces the + and / characters of standard Base64 with - and _ respectively, so that these Base64 signatures no longer need to be URL-encoded.

Therefore this check fails if the key contains a dash (`-`) or underscore (`_`) character as they are not valid in standard base64 encoding:

https://github.com/CartoDB/dataservices-api/blob/402c14e6cabc1c9512d1fa59ab78fd12f0d1be63/server/lib/python/cartodb_services/cartodb_services/google/geocoder.py#L56-L64

(note: that is something that might have changed since it was implemented)

cc/ @ernesmb

rochoa:
cc @juanignaciosl for awareness.

rafatower:
Fixed in https://github.com/CartoDB/dataservices-api/pull/402

PR #402, opened 2017-10-06
GoogleMapsClientFactory to persist clients #401

### LINKED ISSUE #401: google geocoder performs one connection/row
At present, our geocoder is establishing one connection per geocoded row.

This is the current performance bottleneck. Persisting the client between requests will improve its speed in a factor of 2 to 5.

I checked it with the following patch:
```patch
diff --git a/server/lib/python/cartodb_services/cartodb_services/google/geocoder.py b/server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
index e081bd0..9d03772 100644
--- a/server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
+++ b/server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
@@ -17,7 +17,26 @@ class GoogleMapsGeocoder:
         self.client_secret = client_secret
         self.geocoder = googlemaps.Client(
             client_id=self.client_id, client_secret=self.client_secret)
-        self._logger = logger
+
+        import logging
+
+        try:
+            from http.client import HTTPConnection # py3
+        except ImportError:
+            from httplib import HTTPConnection # py2
+        HTTPConnection.debuglevel = 1
+
+        general_logger = logging.getLogger()
+        general_logger.setLevel(logging.DEBUG)
+        handler = logging.FileHandler('/tmp/requests.log')
+        formatter = logging.Formatter('%(asctime)-15s %(message)s')
+        handler.setFormatter(formatter)
+        handler.setLevel(logging.DEBUG)
+        general_logger.handlers = []
+        general_logger.addHandler(handler)
+
+        self._logger = general_logger
+        self._logger.debug("in google geocoder")
 
     def geocode(self, searchtext, city=None, state=None,
                 country=None):
```

Note: fixing https://github.com/CartoDB/dataservices-api/pull/400 is a requirment for keepalive connections.

rafatower:
This fixes #401 

This adds an in-memory cache for clients. Internally they use `Session` objects that persist connections. See https://github.com/googlemaps/google-maps-services-python/pull/164 https://github.com/googlemaps/google-maps-services-python/issues/163 and http://docs.python-requests.org/en/master/user/advanced/#session-objects

As a bonus, it also fixes client keys validity check #382

### DIFF
--- a/server/lib/python/cartodb_services/cartodb_services/google/client_factory.py
+++ b/server/lib/python/cartodb_services/cartodb_services/google/client_factory.py
@@ -10,13 +10,14 @@ class GoogleMapsClientFactory():
 
     @classmethod
     def get(cls, client_id, client_secret):
-        client = cls.clients.get(client_id)
+        cache_key = "{}:{}".format(client_id, client_secret)
+        client = cls.clients.get(cache_key)
         if not client:
             cls.assert_valid_crendentials(client_secret)
             client = googlemaps.Client(
                 client_id=client_id,
                 client_secret=client_secret)
-            cls.clients[client_id] = client
+            cls.clients[cache_key] = client
         return client
 
     @classmethod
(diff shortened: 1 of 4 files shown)


rafatower:
FYI @juanignaciosl @javitonino @ethervoid I'm preparing the release of this along with #400 today and will deploy it on Monday. Cheers!

PR #403, opened 2017-10-10
Number of geometries for OBS_GetAvailableGeometries

ethervoid:
Added number of geometries parameter to the `OBS_GetAvailableGeometries` function in order to be used to improve the geometries `score` calculation 

This change is graceful and backward compatible


client/renderer/interface.yaml
```diff
@@ -427,6 +427,7 @@
     - { name: numer_id, type: text, default: 'NULL' }
     - { name: denom_id, type: text, default: 'NULL' }
     - { name: timespan, type: text, default: 'NULL'}
+    - { name: number_geometries, type: integer, default: 'NULL'}
```
javitonino: Should this be `default: 3000`?

server/extension/sql/130_data_observatory_helper.sql
```diff
@@ -61,10 +61,11 @@ CREATE OR REPLACE FUNCTION cdb_dataservices_server.OBS_GetAvailableGeometries(
   filter_tags TEXT[] DEFAULT NULL,
   numer_id TEXT DEFAULT NULL,
   denom_id TEXT DEFAULT NULL,
-  timespan TEXT DEFAULT NULL)
+  timespan TEXT DEFAULT NULL,
+  number_geometries INTEGER DEFAULT NULL)
```
javitonino: Same question here.

client/renderer/interface.yaml
ethervoid: I'll reply here for both questions.

In the public API I wouldn't set a default for an arbitrary number that doesn't make any sense. I'll leave NULL if the user is not going the scoring functionality.

Maybe what we have to do is to don't calculate or return 0 for the score when the number of geometries is not passed to the observatory-extension function but I'll left as NULL in DS as long as it is a non-mandatory option

Issue #406, opened 2017-11-02
Add parallel and volatile categories to PG functions 


Algunenano:
As done in other projects (for example https://github.com/CartoDB/crankshaft/issues/181) we should add these categories to help the planner make better decisions.

pramsey:
Yes, good stuff.

pramsey:
Also try and find some use cases where you're pretty sure they should parallel and then see what it takes to *force* them parallel. If you can jimmy with costs and get to parallel, that's "good". If nothing you can do gets things parallel, there may be a mis-marked "unsafe" function hiding in there somewhere.

Algunenano:
Some thoughts from the WIP: 
Most, if not all, functions are technically UNSAFE and VOLATILE as they connect and write in Redis (quota management) but, on the other hand if I do something like this:
```sql
SELECT the_geom from mytable WHERE "the_geom" && cdb_geocode_street_point('One street', 'city', 'state', 'country');
```
I'd expect only one call to `cdb_geocode_street_point` to be made. It needs testing, but I think that to have that behaviour the function needs to be at least STABLE. IMMUTABLE is out of the question here (external service and all).

Also, setting this as PARALLEL SAFE could mean several simultaneous calls to the service (here/google/mapzen) which could be an issue if there is any rate limit. RESTRICTED could be OK.

cc/ @rafatower

rafatower:
> I'd expect only one call to cdb_geocode_street_point to be made. It needs testing, but I think that to have that behaviour the function needs to be at least STABLE.

I agree

> Also, setting this as PARALLEL SAFE could mean several simultaneous calls to the service (here/google/mapzen) which could be an issue if there is any rate limit. 

Yes. It's a potential improvement but should be done with care. Some open questions: can we restrict the throughput? at what level/granularity? is the current setup (DS servers) able to cope with the load? Are the external services able too? I think these are good reasons to keep it out of scope FTM and get back to it later on.

My 2 cents

Algunenano:
Some differences on how the planner works depending on the function volatility:

#### VOLATILE
```sql
CREATE OR REPLACE FUNCTION geofunction_volatile (searchtext text)
RETURNS Geometry AS $$
DECLARE
BEGIN
  RAISE NOTICE 'Volatile called';
  RETURN '0101000020E6100000D5E76A2BF69961C00B24287E8C094E40'::geometry;
END;
$$ LANGUAGE 'plpgsql' VOLATILE;
```
[...]

PR #408, opened 2017-11-06
Client: Add PARALLEL and VOLATILE categories

Algunenano:
WIP: Partial implementation for #406 (client functions)


client/renderer/templates/30_plproxy_functions.erb
```diff
@@ -6,4 +6,4 @@ RETURNS <%= return_type %> AS $$
   <% else %>
   SELECT <%= DATASERVICES_SERVER_SCHEMA %>.<%= name %> (<%= params(_with_user_org=true).join(', ') %>);
   <% end %>
-$$ LANGUAGE plproxy;
+$$ LANGUAGE plproxy VOLATILE PARALLEL UNSAFE;
```
jgoizueta: I don't quite understand this: we declare the public functions STABLE, but the internal (plproxy) functions used by them are VOLATILE... I understand all of them should be strictly VOLATILE but we've decided to ignore the Redis-statefulness here for convenience of use... but why are all the plproxy functions VOLATILE?

Algunenano: Yes, strictly speaking they should be VOLATILE. The plproxy are VOLATILE because, as they shouldn't be used directly by the user, there was no need to `lie` to improve performance.
The planner does not dive into the function declaration to see if it's truly STABLE and just blindly trust the programmer, so us "lying" is only needed in the external API functions.

client/Makefile
```diff
@@ -50,6 +55,13 @@ SOURCES_DATA = $(wildcard $(SOURCES_DATA_DIR)/*.sql) $(GENERATED_SQL_FILES)
 $(NEW_EXTENSION_ARTIFACT): $(SOURCES_DATA)
 	rm -f $@
 	cat $(SOURCES_DATA_DIR)/*.sql >> $@
+ifeq ($(PG_PARALLEL), 0)
+# Remove PARALLEL in aggregates and functions
+	$(eval TMPFILE := $(shell mktemp /tmp/$(basename $0).XXXXXXXX))
+	$(SED) -e 's/PARALLEL \= [A-Z]*,/''/g' \
```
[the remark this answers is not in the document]
Algunenano: I know, but I think it might be too late for that.

CONTRIBUTING.md
```diff
@@ -5,6 +5,13 @@ The issue tracker is at [github.com/CartoDB/dataservices-api](https://github.com
 
 We love pull requests from everyone, see [Contributing to Open Source on GitHub](https://guides.github.com/activities/contributing-to-open-source/#contributing).
 
+## PostgreSQL
+When adding or modifying PostgreSQL functions make sure that the
+[VOLATILITY](https://www.postgresql.org/docs/current/static/xfunc-volatility.html) and [PARALLEL](https://www.postgresql.org/docs/9.6/static/parallel-safety.html) categories are updated accordingly.
+As PARALLEL labels need to be stripped for incompatible PostgreSQL versions,
+please use _PARALLEL SAFE/RESTRICTED/UNSAFE_ in uppercase so it's handled
+automatically.
+
```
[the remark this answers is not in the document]
Algunenano: Will do

client/renderer/templates/30_plproxy_functions.erb
jgoizueta: :+1: OK, understood, it's just about not lying more than stricly necessary 😛  thanks for the explanation!

Algunenano:
I've updated the CONTRIBUTING file to clarify a little bit. Can you give a look @jgoizueta, please?

PR #409, opened 2017-11-08
Support for multiple parameters at client (such as channel)

juanignaciosl:
Fixes CartoDB/support/issues/1060.

### DIFF
--- a/server/lib/python/cartodb_services/cartodb_services/google/client_factory.py
+++ b/server/lib/python/cartodb_services/cartodb_services/google/client_factory.py
@@ -9,14 +9,15 @@ class GoogleMapsClientFactory():
     clients = {}
 
     @classmethod
-    def get(cls, client_id, client_secret):
-        cache_key = "{}:{}".format(client_id, client_secret)
+    def get(cls, client_id, client_secret, channel=None):
+        cache_key = "{}:{}:{}".format(client_id, client_secret, channel)
         client = cls.clients.get(cache_key)
         if not client:
             cls.assert_valid_crendentials(client_secret)
             client = googlemaps.Client(
                 client_id=client_id,
-                client_secret=client_secret)
+                client_secret=client_secret,
+                channel=channel)
             cls.clients[cache_key] = client
         return client
 
--- a/server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
+++ b/server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
@@ -5,6 +5,7 @@ import googlemaps
 import re
 
 from exceptions import MalformedResult
+from cartodb_services.google.exceptions import InvalidGoogleCredentials
 from client_factory import GoogleMapsClientFactory
 
 
@@ -12,6 +13,8 @@ class GoogleMapsGeocoder:
     """A Google Maps Geocoder wrapper for python"""
 
     def __init__(self, client_id, client_secret, logger):
+        if client_id is None:
+            raise InvalidGoogleCredentials
         self.client_id = self._clean_client_id(client_id)
         self.client_secret = client_secret
         self.geocoder = GoogleMapsClientFactory.get(self.client_id, self.client_secret)
@@ -49,11 +52,5 @@ class GoogleMapsGeocoder:
 
     def _clean_client_id(self, client_id):
         # Consistency with how the client_id is saved in metadata
-        if client_id is None:
-            return ''
-        else:
-            search_result = re.search(r'client=([^&]*).*', client_id)
-            if search_result is None:
-                return client_id
-            else:
-                return search_result.group(1)
+        search_result = re.search(r'(client=)?(?P<client_id>[^&]*).*', client_id)
+        return search_result.groupdict()['client_id']
--- a/server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
+++ b/server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 import googlemaps
-import re
+from urlparse import parse_qs
 
 from exceptions import MalformedResult
 from cartodb_services.google.exceptions import InvalidGoogleCredentials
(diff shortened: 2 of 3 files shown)


server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
```diff
@@ -48,4 +49,11 @@ def _build_optional_parameters(self, city=None, state=None,
 
     def _clean_client_id(self, client_id):
         # Consistency with how the client_id is saved in metadata
-        return client_id.replace('client=', '')
+        if client_id is None:
+            return ''
+        else:
+            search_result = re.search(r'client=([^&]*).*', client_id)
```
ethervoid: I'd use `(client=)?(?P<client_id>[^&]*).*` as regexp so you can get both options with or without `client=` also using named capturing group makes it more readable. To get the named group you can use `search_result.groupdict()['client_id']`

server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
```diff
@@ -48,4 +49,11 @@ def _build_optional_parameters(self, city=None, state=None,
...
+            if search_result is None:
```
ethervoid: If client_id is not None we should use the provided string that's why I suggested a change of regexp

server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
```diff
@@ -48,4 +49,11 @@ def _build_optional_parameters(self, city=None, state=None,
 
     def _clean_client_id(self, client_id):
         # Consistency with how the client_id is saved in metadata
-        return client_id.replace('client=', '')
+        if client_id is None:
```
ethervoid: I move this check in the init and raise a proper exception in this case instead of making the `_clean_client_id` responsible of validation too

server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
juanignaciosl: Yep, that's what my solution does now (just returning `client_id`), but your proposal is much more elegant 👌

server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
ethervoid: I'd go for [this](https://github.com/CartoDB/dataservices-api/blob/2cf4072b210749e20d786dafc0e63bb58f38becf/server/lib/python/cartodb_services/cartodb_services/google/exceptions.py#L5)

ethervoid:
We have to update the `setup.py` file to increase the version following SEMVER

ethervoid:
We have two ways of authenticating using google maps api key and the premium one which is `client_id` and `client_secret`

The premium option has an optional parameter `channel` that lets the user have a more fine-grained usage report. In order to use that parameter we have to pass it using the `channel` key in the [client initialization](https://github.com/googlemaps/google-maps-services-python/blob/master/googlemaps/client.py#L67), for example:

```
googlemaps.Client(client_id=client_id, client_secret=client_secret, channel=channel)
```
So my suggestion is to modify the regexp and convert the cleaning method into a client_id parser to extract what we need

juanignaciosl:
@ethervoid oh, I didn't see that one. Back to development!

juanignaciosl:
Updated. @ethervoid , don't forget talking to me for paired acceptance, please :_)

server/lib/python/cartodb_services/cartodb_services/google/client_factory.py
```diff
@@ -9,14 +9,15 @@ class GoogleMapsClientFactory():
     clients = {}
 
     @classmethod
-    def get(cls, client_id, client_secret):
-        cache_key = "{}:{}".format(client_id, client_secret)
+    def get(cls, client_id, client_secret, channel=None):
+        cache_key = "{}:{}:{}".format(client_id, client_secret, channel)
```
ethervoid: Hum what happens in case the channel is `None`?

server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
```diff
@@ -2,18 +2,24 @@
...
 
 
 class GoogleMapsGeocoder:
     """A Google Maps Geocoder wrapper for python"""
 
     def __init__(self, client_id, client_secret, logger):
-        self.client_id = self._clean_client_id(client_id)
+        if client_id is None:
+            raise InvalidGoogleCredentials
+        arguments = parse_qs(client_id)
```
ethervoid: :heart: sometimes we try so hard to make the things using the hard way when we have the easiest way at hand!

I'll extract the client_id part but is just a suggestion, nothing serious

server/lib/python/cartodb_services/cartodb_services/google/client_factory.py
juanignaciosl:
```
>>> "{}:{}:{}".format('a', 'b', None)
'a:b:None'
```

LGTM, what do you think?

server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
juanignaciosl: > I'll extract the client_id part but is just a suggestion, nothing serious

What do you mean?

ethervoid: I'm referring to the cliend_id pieces extraction to avoid having it in the init method and give that responsibility to a method

server/lib/python/cartodb_services/cartodb_services/google/client_factory.py
ethervoid: Hum, looks good :+1:

server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
juanignaciosl: Ok, parsing of both parameters extracted to a method.

ethervoid:
### Acceptance

- [x] Tested client_id with channel
- [x] Tested without credentials (gave `ValueError: Must provide API key or enterprise credentials when creating client.`)
- [x] Tested without channel
- [ ] Tested without `client_id=`

PR #420, opened 2017-12-27
Python implementation of clients for Mapbox services

antoniocarlon:
- Routing
 - Geocoding
 - Time Matrix
 - Isochrones (based on Time Matrix)
 - Isodistances (based on isochrones)  <-- We will need to discuss this

https://github.com/CartoDB/cartodb-management/issues/5068
https://github.com/CartoDB/cartodb-management/issues/5069
https://github.com/CartoDB/cartodb-management/issues/5070


server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -535,6 +596,23 @@ def _get_mapzen_config(self):
             self._mapzen_geocoder_quota = mapzen_conf['geocoder']['monthly_quota']
             self._mapzen_geocoder_service_params = mapzen_conf['geocoder'].get('service', {})
 
+    def _get_mapbox_config(self):
+        mapbox_conf_json = self._get_conf('mapbox_conf')
+        if not mapbox_conf_json:
+            raise ConfigException('Mapzen configuration missing')
+        else:
+            mapzen_conf = json.loads(mapbox_conf_json)
```
rafatower: I'd rename this to `mapbox_conf` for clarity

server/lib/python/cartodb_services/cartodb_services/metrics/quota.py
```diff
@@ -72,15 +72,18 @@ def check(self):
         if re.match('geocoder_*',
                     self._user_service_config.service_type) is not None:
             return self.__check_geocoder_quota()
-        elif re.match('here_isolines',
+        elif re.match('mapzen_isolines',
                       self._user_service_config.service_type) is not None:
             return self.__check_isolines_quota()
-        elif re.match('mapzen_isolines',
+        elif re.match('mapbox_isolines',
```
rafatower: I don't get these 2 lines. I think we are losing `here_isolines` in the process.

server/lib/python/cartodb_services/cartodb_services/tools/coordinates.py
```diff
@@ -17,5 +20,43 @@ def longitude(self):
...
+    coordinates.append(coordinates[0])  # Close the ring
+    result_coordinates = []
+    for coordinate in coordinates:
+        result_coordinates.append("%s %s" % (coordinate.longitude,
+                                             coordinate.latitude))
+    wkt_coordinates = ','.join(result_coordinates)
+
+    try:
+        sql = "SELECT ST_CollectionExtract(ST_MakeValid(ST_MakePolygon(ST_GeomFromText('LINESTRING({0})', 4326))),3) as geom".format(wkt_coordinates)
+        geometry = plpy.execute(sql, 1)[0]['geom']
```
rafatower: just keep in mind that you're coupling this python code with the database (probably there's no other cleaner way anyway).

server/lib/python/cartodb_services/cartodb_services/tools/country.py
```diff
@@ -0,0 +1,19 @@
+import plpy
+
+
+def country_to_iso3(country):
```
rafatower: I think this is not used and can be deleted?

server/lib/python/cartodb_services/cartodb_services/tools/qps.py
```diff
@@ -0,0 +1,69 @@
...
+                if response is not None and (response.status_code == 429):
+                    self.retry(start_time, attempt_number)
+                else:
+                    raise e
+            attempt_number += 1
+
+    def retry(self, first_request_time, retry_count):
+        elapsed = datetime.now() - first_request_time
+        if elapsed.total_seconds()  > self._retry_timeout:
+            raise TimeoutException()
```
rafatower: Beware that the message from that exception is `Timeout requesting to mapzen server`

It should be either more generic or tailored to the provider. Otherwise debugging could be a nightmare with false clues :stuck_out_tongue:

server/lib/python/cartodb_services/requirements.txt
```diff
@@ -6,6 +6,7 @@ rollbar==0.13.2
 # Dependency for googlemaps package
 requests==2.9.1
 rratelimit==0.0.4
+mapbox
```
rafatower: I'd lock it to the last stable version

server/lib/python/cartodb_services/setup.py
```diff
@@ -10,7 +10,7 @@
 setup(
     name='cartodb_services',
 
-    version='0.15.6',
+    version='0.15.7',
```
rafatower: I'd bump it to `0.16.0` (kind of trying to follow https://semver.org/ )

server/lib/python/cartodb_services/test/test_quota_service.py
```diff
@@ -179,7 +196,7 @@ def __build_geocoder_quota_service(self, username, quota=100,
                                          username, orgname)
         return QuotaService(geocoder_config, redis_connection=self.redis_conn)
 
-    def __build_routing_quota_service(self, username, provider='mapzen',
+    def __build_routing_quota_service(self, username, provider='mapbox',
```
rafatower: I wouldn't replace the original tests but extend them with the new provider.

server/lib/python/cartodb_services/cartodb_services/tools/country.py
antoniocarlon: This is used both for Mapzen and Mapbox implementations. I have moved it from mapzen to a more generic tools directory

server/lib/python/cartodb_services/cartodb_services/tools/qps.py
antoniocarlon: :man_facepalming:

server/lib/python/cartodb_services/test/test_quota_service.py
antoniocarlon: Done. Added tests for isolines quotas too

server/lib/python/cartodb_services/cartodb_services/metrics/quota.py
antoniocarlon: This was a bug indeed

server/lib/python/cartodb_services/cartodb_services/tools/coordinates.py
antoniocarlon: Yes, I'm aware of the coupling. Another problem is with testing but I think that this is the best way to implement this

server/lib/python/cartodb_services/cartodb_services/mapbox/matrix_client.py
```diff
@@ -0,0 +1,72 @@
...
+    def __init__(self, token, logger, service_params=None):
+        service_params = service_params or {}
+        self._token = token
+        self._logger = logger
+
+    def _uri(self, coordinates, profile=DEFAULT_PROFILE):
+        return BASEURI.format(profile=profile, coordinates=coordinates,
+                              token=self._token)
+
+    @qps_retry(qps=1)
```
ethervoid: We have to check this QPS or use the Headers to adjust it dynamically

server/lib/python/cartodb_services/cartodb_services/mapbox/matrix_client.py
```diff
@@ -0,0 +1,72 @@
...
+        validate_profile(profile)
+        validate_coordinates(coordinates,
+                             NUM_COORDINATES_MIN, NUM_COORDINATES_MAX)
+
+        coords = marshall_coordinates(coordinates)
+
+        uri = self._uri(coords, profile)
+        response = requests.get(uri)
+
+        if response.status_code == requests.codes.ok:
```
ethervoid: Again, take into account possible timeouts

server/lib/python/cartodb_services/cartodb_services/mapbox/routing.py
```diff
@@ -0,0 +1,109 @@
...
+    def _validate_profile(self, profile):
+        if profile not in VALID_PROFILES:
+            raise ValueError('{profile} is not a valid profile. '
+                             'Valid profiles are: {valid_profiles}'.format(
+                                 profile=profile,
+                                 valid_profiles=', '.join(
+                                     [x for x in VALID_PROFILES])))
+
+    def _parse_routing_response(self, response):
+        json_response = json.loads(response)
```
ethervoid: Maybe we should consider to check for empty responses which is going to lead in exception in the line 66

server/lib/python/cartodb_services/cartodb_services/mapbox/routing.py
```diff
@@ -0,0 +1,109 @@
...
+        validate_coordinates(waypoints, NUM_WAYPOINTS_MIN, NUM_WAYPOINTS_MAX)
+
+        coordinates = marshall_coordinates(waypoints)
+
+        uri = self._uri(coordinates, profile)
+        response = requests.get(uri)
+
+        if response.status_code == requests.codes.ok:
+            return self._parse_routing_response(response.text)
+        elif response.status_code == requests.codes.bad_request:
```
ethervoid: In other services you're not checking this to return an empty value. I was tempted to ask for it but we have the `exception safe` functions in the client that returns NULL.

We have two options:

1. Don't check for 4xx errors and return exceptions
2. Do it but then we should keep doing in all the services

The main reason for this code was prior to the exception safe functions when we don't want to cut the geocoding or whatever process due to a 4xx error in one row.

server/lib/python/cartodb_services/cartodb_services/mapbox/matrix_client.py
```diff
@@ -0,0 +1,72 @@
...
+        return BASEURI.format(profile=profile, coordinates=coordinates,
+                              token=self._token)
+
+    @qps_retry(qps=1)
+    def matrix(self, coordinates, profile=DEFAULT_PROFILE):
+        validate_profile(profile)
+        validate_coordinates(coordinates,
+                             NUM_COORDINATES_MIN, NUM_COORDINATES_MAX)
+
+        coords = marshall_coordinates(coordinates)
```
ethervoid: One thing I would like to do is to pass coordinate data objects everywhere instead of strings

server/lib/python/cartodb_services/cartodb_services/mapbox/types.py
```diff
@@ -0,0 +1,4 @@
+MODE_TO_MAPBOX_PROFILE = {
```
ethervoid: The naming is a bit confusing. Maybe something like `TRANSPORT_MODE_TO_MAPBOX`?

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -156,6 +159,8 @@ def __init__(self, redis_connection, db_conn, username, orgname=None):
     def service_type(self):
         if self._routing_provider == self.MAPZEN_PROVIDER:
             return 'routing_mapzen'
+        elif self._routing_provider == self.MAPBOX_PROVIDER:
```
ethervoid: We should change the SQL functions to provide option to use Mapzen or Mapbox if we want to deploy it without start using Mapbox

PR #421, opened 2018-01-04
updating content as per docs issue 1266, ready for review but do not …

csobier:
…merge until given notice.

@inigomedina, new documentation requirements as per Docs issue https://github.com/CartoDB/docs/issues/1266. Please confirm everything is correct but do not merge until we get the okay from management.

*Also, there is some related code in this repo that you may want the developers to take a look at?


ethervoid:
@inigomedina since @csobier is no longer working with us how do we have to proceed with this changes? closing them?

inigomedina:
No, we need to merge it now we are already with Mapbox.

PR #440, opened 2018-01-31
Improve empty requests treatment in geocoders

ethervoid:
* Fix for Mapbox geocoder to handle empty requests and empty responses
* Remove raising an exception when non parameters are passed to the HERE geocoder
* Fix for HERE geocoder with non empty requests
* Added more coverage to the google geocoder credentials parse logic

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 January 29th, 2018
 ==================
+* Version `0.16.3` of the python library
+    * Fix for Mapbox geocoder to handle empty requests and empty responses
+
+January 29th, 2018
+==================
 * Version `0.30.1` of server side
     * Fix for Mapbox geocoding function due to the iso3166 library doesn't support UTF-8 names for the countries
 * Version `0.16.2` of the python library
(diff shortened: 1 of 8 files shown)


server/lib/python/cartodb_services/cartodb_services/mapbox/geocoder.py
```diff
@@ -60,7 +60,10 @@ def _extract_lng_lat_from_feature(self, feature):
     @qps_retry(qps=10)
     def geocode(self, searchtext, city=None, state_province=None,
                 country=None):
-        address = [searchtext]
+        if not searchtext:
```
antoniocarlon: we could add a check for `searchtext.strip()` here

NEWS.md
```diff
@@ -1,3 +1,11 @@
+January 29th, 2018
```
antoniocarlon: We could modify the date here

server/lib/python/cartodb_services/cartodb_services/mapbox/geocoder.py
ethervoid: 👍

ethervoid:
@antoniocarlon time for a second round :)

Issue #455, opened 2018-02-09
Use Mapbox permanent geocoder


antoniocarlon:
As asked by @hannahblue we need to use the Mapbox permanent geocoder instead of the ephemeral one.

PR #456, opened 2018-02-09
Use Mapbox permanent geocoder

antoniocarlon:
Use the Mapbox permanent geocoder instead of the ephemeral one

Closes https://github.com/CartoDB/dataservices-api/issues/455


antoniocarlon:
*Note:* tests are failing because of https://github.com/CartoDB/dataservices-api/issues/457

PR #465, opened 2018-02-13
457 fix mapbox python tests

### LINKED ISSUE #457: Fix Mapbox Python tests
Currently, we are using a Mapbox access token from the Mapbox examples to execute the tests. It seems that they have recently disabled it and our tests are failing.

We don't want to create a CARTO owned access token for the tests because it will be public in our dataservices-api GitHub repo.

@ethervoid suggested that using Travis we can encrypt the config part and decrypt in Travis to use it in the tests (we can't use a configuration file generated with Chef for the tests)

juanignaciosl:
This closes #457. For tests to pass, a `MAPBOX_API_KEY` environment variable must be provided.

### DIFF
--- a/server/lib/python/cartodb_services/test/credentials.py
+++ b/server/lib/python/cartodb_services/test/credentials.py
@@ -1,0 +1,6 @@
+import os
+
+def api_key():
+    """Returns Mapbox API key. Requires setting MAPBOX_API_KEY environment variable."""
+    return os.environ['MAPBOX_API_KEY']
+
--- a/server/lib/python/cartodb_services/test/credentials.py
+++ b/server/lib/python/cartodb_services/test/credentials.py
@@ -1,6 +1,6 @@
 import os
 
-def api_key():
+def mapbox_api_key():
     """Returns Mapbox API key. Requires setting MAPBOX_API_KEY environment variable."""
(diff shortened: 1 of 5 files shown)


juanignaciosl:
Retest this, please

juanignaciosl:
Retest this, please.

server/lib/python/cartodb_services/test/credentials.py
```diff
@@ -0,0 +1,6 @@
+import os
+
+def api_key():
```
ethervoid: I'd call it mapbox_api_key or include mapbox in the file/directory

juanignaciosl: Method rename 👍 This was a remanent of a first iteration that involved a class for namespacing.

server/lib/python/cartodb_services/test/test_mapboxgeocoder.py
```diff
@@ -2,8 +2,8 @@
 from mock import Mock
 from cartodb_services.mapbox import MapboxGeocoder
 from cartodb_services.tools.exceptions import ServiceException
+import credentials
```
ethervoid: Pythonic way `from credential import mapbox_api_key` instead of import everything

Issue #467, opened 2018-02-21
Dataservices legacy mapzen code


jvillarf:
## Description
**Product**: Onpremises `carto-builder-2.2.0 / carto-dataservices-1.1.0`
**Related ticket**: https://github.com/CartoDB/onpremises/issues/455#issuecomment-367329791

Onpremises doesn't configure any dataservices **provider** by default, because that operation is done using onpremises tools. Dataservices **server extension** (`0.30.2`) is falling back to **mapzen** by default for geocoding/isolines/routing.

Additionally, since **mapzen** is officially deprecated, dataservices shouldn't complain if `mapzen_conf` key is not present in `cdb_conf` table.

cc/ @ethervoid

jvillarf:
Hi @ethervoid @antoniocarlon ,

i've been running a couple of tests in an onpremises installation using:
- dataservices_api client extension: `repo tag python-0.17.0`, **0.23.0** version
- dataservices_api server extesion: `repo tag python-0.17.0`, **0.30.2** version

Tests:
- https://github.com/CartoDB/onpremises/issues/466#issuecomment-368853688
- https://github.com/CartoDB/onpremises/issues/466#issuecomment-368857566

Long story short:
- Default fallback to **mapbox** works properly for geocoder and isolines/matrix, but it fails for routing.
   - https://github.com/CartoDB/onpremises/issues/466#issuecomment-368853688
- After defining explicitly **mapbox** as the **routing provider** for a given **organization**, dataservices server is still complaining about **mapzen cdb conf keys** missing.
   - https://github.com/CartoDB/onpremises/issues/466#issuecomment-368857566

I'm not 100% sure if i am missing any configuration step, but so far, the issues point to **routing** as the culprit, because **geocoder and isolines**  are working properly.

In case it helps, the **routing functions** we are using for testing are:
- `cdb_route_with_waypoints`
- `cdb_route_point_to_point`

PR #468, opened 2018-02-21
Remove legacy mapzen

ethervoid:
Related #467


server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -589,8 +589,9 @@ def _get_here_config(self):
 
     def _get_mapzen_config(self):
         mapzen_conf_json = self._get_conf('mapzen_conf')
+        # We dont use mapzen anymore so we don't need to check for its configuration
         if not mapzen_conf_json:
-            raise ConfigException('Mapzen configuration missing')
+            return
         else:
```
antoniocarlon: Why not just `if mapzen_conf_json:` instead of the `if - else`?
It's OK now if you want to keep the same structure as the other providers and point out that we are not raising an exception so it's up to you

PR #471, opened 2018-02-26
Fix Mapbox isolines near the shoreline

antoniocarlon:
Using the returned destination coordinates from the Mapbox matrix client as the isolines coordinates instead of the calculated ones.

The complete solution would be actually clipping the returned polygon using the shoreline, but this may be a good enough approximation.

Old behavior:
screenshot from 2018-02-26 12-59-39

New behavior:
screenshot from 2018-02-26 13-00-39

**Note:** as a side effect, as we are using the Mapbox returned coordinates instead of the calculated ones, we are slightly improving the rest of the isoline (notice the change in the borders).

Fixes https://github.com/CartoDB/support/issues/1333

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,9 @@
 
+February 22th, 2018
+==================
+* Version `0.17.2` of the python library
+    * Fix bug with Mapbox isolines not stopping at the seacoast
+
 February 27th, 2018
 ==================
 * Version `0.17.1` of the python library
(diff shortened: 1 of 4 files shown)


juanignaciosl:
Before (same parameters than the ones from the issue):

captura de pantalla 2018-03-01 a las 13 21 54

After (2500 instead of 2700 seconds as a way to avoid the cache):

captura de pantalla 2018-03-01 a las 13 21 17

Is that glitch something that we could avoid?

antoniocarlon:
Removing the points that are farther away (in time) than the given isorange plus the selected tolerance seems to do the trick and now we get pretty neat isolines near the shore.

juanignaciosl:
Ferpect!

2700:

captura de pantalla 2018-03-05 a las 12 59 00

2500:

captura de pantalla 2018-03-05 a las 13 01 07

PR #472, opened 2018-02-27
Fix mapzen routing bug

ethervoid:
Related #467 

We keep trying to gather Mapzen credentials from the database but now is possible to not have them so that leads to an exception


jvillarf:
After running some tests (https://github.com/CartoDB/onpremises/issues/466), **default routing provider** is now **mapbox**. Thanks a lot.

PR #476, opened 2018-03-13
Avoid reaching provider for empty geocodings

antoniocarlon:
Avoid reaching provider for empty geocodings (but still incrementing empty service use)


juanignaciosl:
Uhm, tests failed because of extension installation:

```
   8   CREATE EXTENSION cdb_geocoder;
   9 + ERROR:  could not open extension control file "/usr/share/postgresql/10/extension/cdb_geocoder.control": No such file or directory
```

[regression.diffs.zip](https://github.com/CartoDB/dataservices-api/files/1807694/regression.diffs.zip)

antoniocarlon:
Yes, we definitely need to work on avoiding the duplication. The milestone of adding a new provider can be a good moment (not far in time, I hope ;))

On the other hand, I'm currently working on the tests to see why they're failing.

juanignaciosl:
👍 about duplication.

Tests: it seems to lack geocoder extension 🤔

antoniocarlon:
Yes, they lack the geocoder extension, but they were also failing in my local installation for a totally different reason and that was confusing me (just wanted to make sure that the tests are passing locally before going to the failing tests here)

skgsergio:
retest this please

skgsergio:
Tests are now working.

ethervoid:
Hard to achieve the de-duplication because we're talking about SQL functions, you can't extract to another method and if you do it has to be a new SQL function or put it inside the python library which we partially have with the `ServiceManager`

Also we have to take into account that some providers need some extra work before returning the SQL result set so it's going to be hard.

That said I don't like it neither and if we find a way to do it perfect but is not going to be an easy one if possible

server/extension/sql/20_geocode_street.sql
```diff
@@ -98,6 +98,11 @@ RETURNS Geometry AS $$
 
   try:
     service_manager.assert_within_limits()
+
+    if not searchtext and not city and not state_province and not country:
+      service_manager.quota_service.increment_empty_service_use()
```
ethervoid: We should move this to the python library and return empty result or none in case the user passes an empty request but before we make a request to the service

antoniocarlon: I don't like the idea of relying on the Python library for these checks. If the SQL functions receive empty data, they should manage it. What do you think?

ethervoid: I think this should go into the python library for multiple reasons:

* We use the SQL functions as a proxy but the real deal is done in the python library which includes sanitization, checks [like this](https://github.com/CartoDB/dataservices-api/blob/development/server/lib/python/cartodb_services/cartodb_services/here/geocoder.py#L77), return empty if the request is not correct, and so on
* It's far easier an reliable to deploy only the python library than change the SQL functions
* Our goal is to move everything that is possible to the python library which is the core for the DS extension: it knows how to check quota, how to gather configuration, how to make a request, how to deal with the returned object, how to check for possible discrepancies and so on
* We can manage far better the duplication inside a python library than in the SQL functions

So for those reasons I prefer to have all the sanity checks inside the function, return an empty result and let the SQL function to be agnostic about it

antoniocarlon: Your explanation made me change my mind... the argument about having the SQL functions as a proxy for the Python code is more than enough to reconsider it.

I'm going to undo the changes and revisit the Mapbox geocoder function and validity checks.

antoniocarlon:
We already had validity checks for Here and Mapbox on the Python side, but the one implemented for Mapbox was not correct (having just the city or the state should be enough for geocoding)

juanignaciosl:
@ethervoid I guess that current code filters on "our side" ("don't call 3rd party libraries if there aren't parameters") while the old one filtered on "their side". I think that this is the correct approach. What do you think?

ethervoid:
From my PoV we're using that geocoding for street level and named places which could be done using the search text part and the cities/country parameters to make the results more accurate but I didn't think in those parameters as a way to get Spain or Valladolid for example.

I don't have a problem merging these changes but I wanted to know your opinion about it

juanignaciosl:
I don't have strong feelings about this, I'd merge.

ethervoid:
Okiiiiii

PR #488, opened 2018-03-23
Cc211 isolines should always return multipolygon

juanignaciosl:
Extension-side of CartoDB/cartodb-central/issues/2151.


juanignaciosl:
To Data Team kanban for CR.

ethervoid:
### Acceptance 

- [x] Tested jtorres example before and after deploy the change to see how it change from returning POLYGON to MULTIPOLYGON
- [x] Tested HERE and Mapbox providers
- [x] Tested downgrade and upgrade of the extension

PR #497, opened 2018-05-22
Bulk geocoding PoC

ethervoid:


server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
```diff
@@ -33,6 +46,23 @@ def geocode(self, searchtext, city=None, state=None,
...
+        results = []
+        decoded_searchtext = json.loads(searchtext)
+        for search in decoded_searchtext:
+            bulk_results.append(pool.apply_async(async_geocoder, (self.geocoder, search)))
+        pool.close()
+        pool.join()
+        try:
+            for bulk_result in bulk_results:
+                result = bulk_result.get()
+                results.append([result[0], self._extract_lng_lat_from_result(result[1][0]), []])
```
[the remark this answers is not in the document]
ethervoid: Yes, if we get some error when we call `get` it's going to fail everything so we should tackle this. In case of hit the rate limit, we're going to receive a 429 which leads to not having results and finally get a failure in the `_extract_lng...` function. Also, I haven't found valuable metadata coming from google results but metadata is very useful to identify problems in the future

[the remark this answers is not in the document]
juanignaciosl: Thanks!

server/lib/python/cartodb_services/cartodb_services/google/geocoder.py
```diff
@@ -33,6 +46,23 @@ def geocode(self, searchtext, city=None, state=None,
         except KeyError:
             raise MalformedResult()
 
+    def bulk_geocode(self, searchtext):
+        pool = Pool(processes=13)
```
jaakla: based on my tests and calcs you should get max performance with parallel processes=100. This is in line with google limit: 5000 req / 100 sec / 0.5 sec/req = 100. You are not restricted here by CPUs, but waiting after google response, so can be aggressive. But some safety may be needed, and consideration that several users may do geocoding in the same time, even if with google it is unlikely.

ethervoid: I've changed it to test and I haven't received any problem. I made a mistake in my calculations but this issue shows the gain of performance and using 13 works is a good approximation. You can see the chart [here](https://github.com/googlemaps/google-maps-services-python/issues/222)

Using 13 works is my try to made a tradeoff between limits

jaakla: Here are my numbers btw, dont have fancy graph. Calculated with https://github.com/CartoDB/geocoder-tests/blob/master/Geocoder-google.ipynb:
* serial 100 geocodes: Wall time: 55.5 s
* parallel 10 workers: Wall time: 13.6 s
* parallel 20: Wall time: 4.24 s
* parallel 50: Wall time: 2.66 s
* 1000 geocodes parallel 100 workers:  Wall time: 24.7 s  Wall time: 29.5 s (2nd run)
* 1000 gc, parallell 10 workers: Wall time: 1min 1s
* 5000 gc, parallell 100 workers: wall time: 2 min

So for me 10->50 gave about 2x extra win in pure test, but of course need to consider whole context here.

PR #501, opened 2018-06-27
Fix mapbox routing fixture

juanignaciosl:


juanignaciosl:
@ethervoid ok to merge?

PR #503, opened 2018-07-02
Geocoder boost here

juanignaciosl:


server/lib/python/cartodb_services/cartodb_services/geocoder.py
```diff
@@ -0,0 +1,76 @@
...
+
+    logger = Logger(logger_config)
+
+    try:
+        service_manager.assert_within_limits(quota=False)
+        geocode_results = geocoder.bulk_geocode(searches=searches)
+        if geocode_results:
+            results = []
+            for result in geocode_results:
+                if result[1]:
```
antoniocarlon: We may want to check that `len(result[1])` is exactly `2`

server/lib/python/cartodb_services/cartodb_services/geocoder.py
```diff
@@ -0,0 +1,76 @@
...
+    finally:
+        service_manager.quota_service.increment_total_service_use()
+
+
+StreetGeocoderSearch = namedtuple('StreetGeocoderSearch', 'id address city state country')
+
+
+class StreetPointBulkGeocoder:
+    """
+    Classes extending StreetPointBulkGeocoder should implement:
```
antoniocarlon: You could add something like
```
def _bulk_geocode(self, street_geocoder_searches):
    """
    Subclasses must implement _bulk_geocode
    """
    raise NotImplementedError('Subclasses must implement _bulk_geocode')
```

server/lib/python/cartodb_services/cartodb_services/here/bulk_geocoder.py
```diff
@@ -0,0 +1,153 @@
...
+HereJobStatus = namedtuple('HereJobStatus', 'total_count processed_count status')
+
+class HereMapsBulkGeocoder(HereMapsGeocoder, StreetPointBulkGeocoder):
+    BATCH_URL = 'https://batch.geocoder.cit.api.here.com/6.2/jobs'
+    MAX_BATCH_SIZE = 1000000 # From the docs
+    # https://developer.here.com/documentation/batch-geocoder/topics/read-batch-request-output.html
+    META_COLS = ['relevance', 'matchType', 'matchCode', 'matchLevel', 'matchQualityStreet']
+    MAX_STALLED_RETRIES = 100
+    BATCH_RETRY_SLEEP_S = 5
+    MIN_BATCHED_SEARCH = 100 # Under this, serial will be used
```
antoniocarlon: `MAX_STALLED_RETRIES` and `MIN_BATCHED_SEARCH` are both `100`. Is this intended?

server/lib/python/cartodb_services/cartodb_services/here/bulk_geocoder.py
```diff
@@ -0,0 +1,153 @@
...
+        if len(searches) > self.MAX_STALLED_RETRIES:
+            raise "Batch size can't be larger than {}".format(self.MAX_STALLED_RETRIES)
+        if self._should_use_batch(searches):
+            self._logger.debug('--> Batch geocode')
+            return self._batch_geocode(searches)
+        else:
+            self._logger.debug('--> Serial geocode')
+            return self._serial_geocode(searches)
+
+    def _should_use_batch(self, searches):
```
antoniocarlon: Are we expecting this logic to be very different between geocoders? Could we extract this to the parent class and let the subclasses just implement `_serial_geocode` and `_batch_geocode` and the parameters?

server/lib/python/cartodb_services/cartodb_services/here/bulk_geocoder.py
```diff
@@ -0,0 +1,153 @@
...
+        return results
+
+    def _searches_to_csv(self, searches):
+        queue = cStringIO.StringIO()
+        writer = csv.writer(queue, delimiter='|')
+        writer.writerow(['recId', 'searchText', 'country'])
+
+        for search in searches:
+            fields = [search.address, search.city, search.state]
+            search_text = ', '.join(filter(None, fields))
```
antoniocarlon: The examples (https://developer.here.com/api-explorer/rest/batch_geocoding) show the address component separated using a white space instead of a comma (just to raise the notice)

server/lib/python/cartodb_services/cartodb_services/here/bulk_geocoder.py
```diff
@@ -0,0 +1,153 @@
...
+                                    params=self.credentials_params,
+                                    timeout=(self.connect_timeout, self.read_timeout))
+        root_zip = zipfile.ZipFile(io.BytesIO(result_r.content))
+
+        results = []
+        for name in root_zip.namelist():
+            if name.endswith('_out.txt'):
+                reader = csv.DictReader(root_zip.open(name), delimiter='|')
+                for row in reader:
+                    if row['SeqNumber'] == '1':  # First per requested data
```
antoniocarlon: Can we expect the results to be ordered by relevance/quality?

test/integration/test_street_functions.py
```diff
@@ -54,11 +100,11 @@ def test_full_spec(self):
...
-        assert_equal(response['total_rows'], 2)
-
-        row_by_cartodb_id = self._row_by_cartodb_id(response)
-        self._assert_x_y(row_by_cartodb_id[1], -4.728252, 41.6517025);
-        self._assert_x_y(row_by_cartodb_id[2], -4.7404453, 41.6314339)
+        points_by_cartodb_id = {
+            1: self.fixture_points['Plaza Mayor, Valladolid'],
+            2: self.fixture_points['Paseo Zorrilla, Valladolid']
+        }
+        assert_equal(self._x_y_by_cartodb_id(response), points_by_cartodb_id)
```
antoniocarlon: The results from the geocoder can change as the provider refine the algorithms. We could round the results to get less spurious errors

server/lib/python/cartodb_services/cartodb_services/geocoder.py
juanignaciosl: Not sure about this. First of all, you probably mean `len(result)`, because `result[1]` is the name of the column (if there's a match with the `{{ ... }}` pattern), or `undefined` (if there isn't).

That said, I'm not sure about which case might lead to a case where `len` were bigger. Even `{{ cartodb_id  wadus }}` (which BTW is wrong) leads to the same size.

Which scenario are you thinking of?

server/lib/python/cartodb_services/cartodb_services/here/bulk_geocoder.py
juanignaciosl: Nop, just coincidence. `MIN_BATCHED_SEARCH` value intention is more accurate, because after a few tests the bulk improvement seems to be around that number. `MAX_STALLED_RETRIES` just happens to be 100. I don't have evidence for a good number, so I chose a high one to avoid errors if HERE has a glitch.

server/lib/python/cartodb_services/cartodb_services/here/bulk_geocoder.py
juanignaciosl: I don't know yet. Google is a direct win because it's just parallelization of the same API. HERE, on the other hand, isn't. I'd defer this decision until more providers are implemented.

server/lib/python/cartodb_services/cartodb_services/geocoder.py
antoniocarlon: `result[1]` is the array containing the longitude and latitude for the `ST_MakePoint`. An error decoding the geocoder result could lead to this array to be shorter

server/lib/python/cartodb_services/cartodb_services/here/bulk_geocoder.py
juanignaciosl: [Other examples contain commas](https://developer.here.com/documentation/batch-geocoder/topics/data-input.html). I've been searching for the official recommendation, and there seems to be none, so I'd keep commas. Nice catch, though!

server/lib/python/cartodb_services/cartodb_services/here/bulk_geocoder.py
antoniocarlon: Yes, but I mean that for `_should_use_batch` to return `True`,  `len(searches) >= self.MIN_BATCHED_SEARCH` (>=100), and if `len(searches) > self.MAX_STALLED_RETRIES`, it would raise the `"Batch size can't be larger than {}"`

server/lib/python/cartodb_services/cartodb_services/here/bulk_geocoder.py
juanignaciosl: I see no explicit mention to it in the documentation. In the tests that I've run, they are.

server/lib/python/cartodb_services/cartodb_services/here/bulk_geocoder.py
juanignaciosl: Oh, true. It was using ` MAX_STALLED_RETRIES` for the message, but it should be `MIN_BATCHED_SEARCH`.

juanignaciosl:
@antoniocarlon ready for the next CR

test/integration/test_street_functions.py
antoniocarlon: Sounds good, but we may want to set a different tolerance for `assert_close_enough` depending on the geocoder (0.0005 degrees is about 50 meters at the equator)

juanignaciosl: :+1: I'll address it with the third provider.

juanignaciosl:
I think that all comments are addressed. @antoniocarlon when this is :+1: for you, merge it into `geocoder_boost` and CartoDB/dataservices-api/pull/499 and CartoDB/camshaft/pull/367 will be ready for acceptance.

PR #511, opened 2018-08-02
Dataservices dependencies specification

juanignaciosl:
Related: CartoDB/onpremises/issues/570

### DIFF
--- a/server/extension/carto-package.json
+++ b/server/extension/carto-package.json
@@ -1,30 +1,14 @@
 {
   "name": "dataservices-api-server-extension",
-  "versions": {
-    "0.32.0": {
-      "tag": "0.32.0-server",
-      "works_with": {
-        "dataservices-api-server-python-lib": "^0.19.0"
-      },
-      "requires": {
-        "postgresql": "^10.0.0",
-        "postgis": "^2.4",
-        "cartodb-postgresql": "^0.23.0"
-      }
+  "current_version": {
+    "requires": {
+      "postgresql": "^10.0.0",
+      "postgis": "^2.4",
+      "cartodb-postgresql": "^0.23.0"
     },
-    "0.32.0-customer": {
-      "tag": "0.32.0-server-customer",
-      "works_with": {
-        "dataservices-api-server-python-lib": {
-          "git-repository": "https://github.com/CUSTOMER/dataservices-api",
-          "manifest": "/server/extension/carto-package.json",
-          "version": "0.19.0-experimental"
-        }
-      },
-      "requires": {
-        "postgresql": "^10.0.0",
-        "postgis": "^2.4",
-        "cartodb-postgresql": "^0.23.0"
+    "works_with": {
+      "dataservices-api-server-python-lib": "^0.19.1",
+      "observatory-server-extension": "^1.9.0"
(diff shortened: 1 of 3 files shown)


server/lib/python/cartodb_services/carto-package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "dataservices-api-server-python-lib",
+  "versions": {
+    "0.19.0": {
```
juanignaciosl: This is the closest example to your proposal, just with the `versions` addition, let me explain it: 

For repos with more than one component, we can't have the version <-> tag assumption, because there would be collisions. We could workaround it with tag templating (`python-VERSION` in our case), but IMHO it's complicated. So, we need a way to match versions with tags. Doing this have nice side effects, such as supporting more complex scenarios (more about this at server/extension/carto-package.json comment).

`tag` can be optional, assuming that it's the `version` name.

server/extension/carto-package.json
```diff
@@ -0,0 +1,30 @@
...
+      "requires": {
+        "postgresql": "^10.0.0",
+        "postgis": "^2.4",
+        "cartodb-postgresql": "^0.23.0"
+      }
+    },
+    "0.32.0-customer": {
+      "tag": "0.32.0-server-customer",
+      "works_with": {
+        "dataservices-api-server-python-lib": {
```
juanignaciosl: Instead of a version, a dependency could be a complete specification. This enables definition for customers that have forked repos, for example (we have cases like this), and also makes the "carto-package specification" self-contained: the root file is a normal "carto-package", not an exception.

`manifest` attribute is needed for repositories where the file is not at the root. If not present, you could assume that it's at the root.

lekum:
cc @skgsergio for your awareness :wink: 

Thank you very much @juanignaciosl :clap: :clap:

skgsergio:
Thanks, I think the scheme you propose makes sense if we have at least one case like this.

We now have to think about this proposal and how affects the whole idea. Thanks for the effort as this is something we wanted with the meeting, try to identify issues like this 👍

lekum:
Hi @juanignaciosl 

@skgsergio and I have examined your proposal and looks reasonable, so we are going to propose it to the attendees of the project meeting in order to check their opinions and look for a consensus.

On our view, there are tradeoffs. Maintaing a long list of versions can make the `carto-package.json` a bit cumbersome, but on the other side, having the list of versions means that we only have to do one checkout per repo to get all the version restrictions.

There are some assumptions that we need to make in order for this approach to work

- The version in `master` of the `carto-package.json` should be the source of truth. This approach become exponentially more difficult if we need to checkout the tags related to every version in the repo to get the list of versions again. So we should assume that the `carto-package.json` file(s) in `master` are the authoritative ones.

- Although we have now decoupled the tag format from the version number, we should use semantic versioning in the version keys in the `carto-package.json` in order to be able to enforce the restrictions, calculate compatible versions, etc...

- We still don't know if we are going to model the `onpremises` version as a `carto-package.json` file, but we will somehow have the configuration in the repo to search for the `carto-package.json` of each component in the `master` branch and thus start the compatibility search.

PR #515, opened 2018-08-20
carto-package.json

juanignaciosl:
@lekum after #514 you probably want this to be available in master. Just merge for confirmation :-) !


lekum:
Thank you very much! :+1:

Issue #518, opened 2018-08-31
Use dataservices with non-master keys


javitonino:
It's finally time to talk about the issue with dataservices not being able to be executed with non-master API Keys. It's something that we want to do for the OAuth implementation to be useful. The problem is as follows. SQL API received the API Key and executed the DS service with the matching API Key role. But the API Key is not passed deeper into the stack. Instead, DS matches the role name against some rules to determine the username. This works for the naming of roles in master keys, but not for others.

More discussion: https://github.com/CartoDB/cartodb-management/issues/5191

Objectives:
- A way to use DS without master keys
- Be able to implement permissions. What makes sense here? Maybe the basics (geocoding/isolines/DO)?

antoniocarlon:
The API key data would be like this:
```
{
	"username": "myusername",
	"organization": "",
	"application": "myapplication",
	"permissions": {"dataservices": ["geocoding", "routing", "isolines"]},
	"quotas": {"dataservices": ["geocoding": 100, "isolines": 100]},  <----------- TBD
	"rate_limit": {"dataservices": ["isolines": 5]}  <----------- TBD
}
```

We have several options for the API keys/configuration storage:
**Using `cdb_conf`:**
 - Creating a key `api_keys` to store all the API keys and their permissions:
```
key = api_keys
value = {
	"database_role1": {
		// ...
	}, 
	"database_role2": {
		// ...
	}, 
}
```
The main drawback is that all the API keys would be stored under the same key, and this can be especially problematic for organizations because all the API keys for all the users would be stored here (in a huge JSON)
 - Creating a key `database_role` (maybe with a prefix?) to store the permissions of the API key related to that role:
```
key = database_role
value = // the API key data from above (JSON)
```
The main drawback is that the current organization of `cdb_conf` doesn't seem to favour this approach
**Using a new table:**
Create a new table with the following schema:
```
role | username | organization | application | permissions
```

cc @oleurud  @javitonino @juanignaciosl 

Note: We have already talked about this (using `cdb_conf`), but when we were defining the structure we have found these issues and we would like to discuss it a bit more

javitonino:
> `"organization": ""`

Do we need this? It should be in an entry `cdb_conf` already (since the organization is the same for the whole database).

About the three options:
1. Single key. I would discard it because it makes for slower lookups/updates than the alternatives.
[...]

PR #519, opened 2018-09-06
Use non-master keys

antoniocarlon:
- Added permissions in `cdb_conf` and check them in dataservices
 - Added the `appname` parameter to the functions

Closes https://github.com/CartoDB/dataservices-api/issues/518


antoniocarlon:
This is the first working version of the dataservices modifications.

@juanignaciosl @javitonino @oleurud Some open discussions:

 - I have just added the permissions to the geocoding, routing and isolines functions, but not to the segments, or observatory functions. We could discuss whether we want to add permissions to them.
 - I have added the `appname` parameter to the public server functions and to the ones that save metrics, but not to all the private server functions (all of them receive the `username` and `orgname`  and we may want to pass them the `appname` too for symmetry)

**Notes:**
 - The contract for the public server functions have changed to add the `appname`. This means that when we deploy the server, all the queries will fail until we deploy the client
 - I haven't created the upgrade/downgrade/create extension scripts yet (neither for the client nor for the server). I'll wait until we agree that this is a good version to start working with

client/renderer/templates/20_public_functions.erb
```diff
@@ -9,11 +9,17 @@ RETURNS <%= return_type %> AS $$
 DECLARE
   <%= return_declaration if not multi_row %>
   <%= user_org_declaration %>
+  <% unless superuser_function? -%>appname text;
```
javitonino: Going full YAGNI here. We want to know that this is possible since we are going to be implemeting it soon. Since we are still not using it for anything, I think we can remove it for now. But it's important to know it can be easily added.

Also: would it be possible to set a default for this parameter? That way we should be compatible with older client releases.

client/sql/15_config_management.sql
```diff
@@ -1,6 +1,8 @@
 CREATE TYPE cdb_dataservices_client._entity_config AS (
     username text,
-    organization_name text
+    organization_name text,
+    application_name text,
```
javitonino: YAGNI

client/sql/15_config_management.sql
```diff
@@ -13,10 +15,13 @@ CREATE OR REPLACE FUNCTION cdb_dataservices_client._cdb_entity_config()
 RETURNS record AS $$
 DECLARE
     result cdb_dataservices_client._entity_config;
+    apikey_config json;
     is_organization boolean;
     username text;
     organization_name text;
 BEGIN
+    SELECT cartodb.cdb_conf_getconf('api_keys_'||session_user) INTO apikey_config;
```
javitonino: What happens if the entry is missing? I'm guessing some kind of error, but I'd like it to be a controlled one (since these errors bubble up to the user).

javitonino: The tests make it clearer, it returns the generic permission missing error. :+1:

client/renderer/templates/20_public_functions.erb
antoniocarlon: With the current implementation we are using the `appname` for the metrics in the server functions. This is related to the questions in the comment: adding the `appname` will break the current contract between the client and the server functions. Of course, we can remove the `appname` here and in the server if we want to go without the application metrics. What do you think?

javitonino: IMHO, we can go without the metrics. That project is ill-defined, so requirements may change that imply that we want to track the specific API Key instead of the app name, etc.

The implementation of metrics is not really in the scope of the project (but thinking of a solution extensible enough to be able to add them was).

In short, let's remove this for now.

antoniocarlon:
@javitonino I have removed the application from the parameters, another CR round?

antoniocarlon:
Sorry, I forgot to remove a couple of occurrences. Now it's ready for a new CR!

client/sql/15_config_management.sql
```diff
@@ -13,25 +14,21 @@ CREATE OR REPLACE FUNCTION cdb_dataservices_client._cdb_entity_config()
...
-        FROM pg_namespace s
-        LEFT JOIN pg_roles r ON s.nspowner = r.oid
-        WHERE r.rolname = session_user INTO username;
         SELECT cartodb.cdb_conf_getconf('user_config')->>'entity_name' INTO organization_name;
-    ELSE
-        SELECT cartodb.cdb_conf_getconf('user_config')->>'entity_name' INTO username;
-        organization_name = NULL;
     END IF;
-    result.username = username;
+    result.username = apikey_config->>'username';
```
javitonino: Just to make sure we are aware, this means we have to deploy Rails changes and run a rake before deploying this.

javitonino:
Acceptance to be done with the Rails part: https://github.com/CartoDB/cartodb/issues/14270

client/sql/15_config_management.sql
antoniocarlon: Yes, I have updated the deployment steps here https://github.com/CartoDB/product/issues/144#issuecomment-418364252

antoniocarlon:
As this is ready for the acceptance, I'm creating the create/upgrade/downgrade scripts

antoniocarlon:
@javitonino, a small CR for the DO permissions? Changes in these files:
 - `client/renderer/interface.yaml` https://github.com/CartoDB/dataservices-api/pull/519/files#diff-8daf1cf575bb2e87d53aa4f82df508d5
 - `client/test/expected/90_data_observatory_test.out` https://github.com/CartoDB/dataservices-api/pull/519/files#diff-044e48674e8cbae43de52bf3ef5f315b
 - `client/test/sql/90_data_observatory_test.sql` https://github.com/CartoDB/dataservices-api/pull/519/files#diff-7fc87041b17df9f61145471ba979079d

juanignaciosl:
Is this still blocked?

javitonino:
Yup. Finishing up cartodb's side of things -> https://github.com/CartoDB/cartodb/pull/14270

javitonino:
Ready to test!

javitonino:
Tested:

- Old DS client with new DS server :white_check_mark:  (let's deploy server before client?)
- New DS client
  - Without running the rake, everything fails with error :white_check_mark: 
  - Running the rake: master key works, public key doesn't work :white_check_mark: 
  - New regular key with some permissions, works for what is granted :white_check_mark: 

:red_circle: When trying to get OBS metadata:
```
cdb_dataservices_client._obs_getavailablenumerators(7): [dataservices_db] REMOTE ERROR: spiexceptions.ExternalRoutineException: ImportError: cannot import name ObservatorySnapshotConfig
```

javitonino:
Uhm, server extension deploy failed without being marked as failure, that's likely to be causing the last problem...

javitonino:
Now it works after manually upgrading :rocket:

PR #521, opened 2018-09-13
Fixed tests

antoniocarlon:
Fixed integration tests.
Related to https://github.com/CartoDB/cartodb/issues/14283

Tests must handle two error messages, since for public api keys in non-org roles (which use postgres built-in `public` role) the DS client schema is not in the search path. See https://github.com/CartoDB/cartodb/issues/14283

### DIFF
--- a/test/integration/test_admin0_functions.py
+++ b/test/integration/test_admin0_functions.py
@@ -1,6 +1,6 @@
 from unittest import TestCase
 from nose.tools import assert_raises
-from nose.tools import assert_not_equal, assert_equal
+from nose.tools import assert_not_equal, assert_true
 from ..helpers.integration_test_helper import IntegrationTestHelper
 
 
@@ -12,7 +12,6 @@ class TestAdmin0Functions(TestCase):
             self.env_variables['schema'],
             self.env_variables['username'],
             self.env_variables['host'],
-            self.env_variables['api_key']
         )
 
     def test_if_select_with_admin0_is_ok(self):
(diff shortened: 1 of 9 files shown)


test/integration/test_admin0_functions.py
```diff
@@ -29,4 +28,4 @@ def test_if_select_with_admin0_without_api_key_raise_error(self):
         try:
             IntegrationTestHelper.execute_query(self.sql_api_url, query)
         except Exception as e:
-            assert_equal(e.message[0], "The api_key must be provided")
+            assert_true(e.message[0] in ["Geocoding permission denied", "function cdb_geocode_admin0_polygon(text) does not exist"])
```
javitonino: I'd user `assert_in` for this:

`assert_in(e.message[0], ["Geocoding permission denied", "function cdb_geocode_admin0_polygon(text) does not exist"])`

Repeat this comment in a few more places :upside_down_face:

javitonino:
I altered your PR comment just in case someone follows the blame commits to this. I added a little explanation of why we are doing this :)

antoniocarlon:
Added suggestion, thank you! ready for the next CR round!

javitonino:
:red_circle: 

Many failures related to:
```
assert_in(e.message[0], ["Observatory permission denied", "function obs_getuscensusmeasure(geometry, unknown) does not exist"])
AssertionError: u'Data Observatory permission denied' not found in ['Observatory permission denied', 'function obs_getuscensusmeasure(geometry, unknown) does not exist']
```

But there are some others:
`test_if_select_with_admin0_without_api_key_raise_error` fails due to not having permissions on the table it wants to geocode

`test_full_spec`, `test_component_aggregation` and `test_templating_geocoding` fail because the point is not close enough (might be a thing of geocoder provider config?)

`Useful just to test a good batch size` fails with a JSON decoding issue.

PR #527, opened 2018-10-18
Include note to explain why some isodistances could not be precise

### LINKED ISSUE #526: Change Isodistance documentation
Related https://github.com/CartoDB/support/issues/1788#issuecomment-429893228

We need to include in the isodistance documentation that the precision of the created isoline depends on the provider.

For example HERE let us pass the desired distance, in meters, of the isoline but TomTom doesn't. With tomtom we have to calculate the amount of maximum time for the isoline in seconds and we have to take into account the limitations of the lanes and so on.

ethervoid:
Closes #526 

@ramiroaznar who could review this doc change?

// @alexsroth

### DIFF
--- a/doc/isoline_functions.md
+++ b/doc/isoline_functions.md
@@ -16,6 +16,8 @@ The following functions provide an isoline generator service, based on time or d
 
 Displays a contoured line on a map, connecting geometries to a defined area, measured by an equal range of distance (in meters).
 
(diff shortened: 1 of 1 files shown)


ramiroaznar:
LGTM. But remember to include also this change on this branch https://github.com/CartoDB/dataservices-api/tree/developer-center in order to see it on the carto.com/developers

PR #528, opened 2018-10-24
Fix qps tomtom

ethervoid:
Related https://github.com/CartoDB/support/issues/1811

That has a great problem when we're dealing with legit 403 status for
example deactivated user, forbidden access, etc.

I've added a check for the HTTP header `X-Error-Detail-Header` in order
to distinguish between legit 403 and 429 error messages

Possible values for `X-Error-Detail-Header` in a 403 error:

  - Service Requires SSL : http is used instead of https (secure)

  - Invalid Referer : invalid 'Referer' header value is send
  to https://api.tomtom.com and allowed referer values are
  configured on specific API key

  - Account Over Queries Per Second Limit : rate limit exceeded

  - Account Inactive : incorrect API key/API key no longer valid

### DIFF
--- a/server/lib/python/cartodb_services/cartodb_services/tools/qps.py
+++ b/server/lib/python/cartodb_services/cartodb_services/tools/qps.py
@@ -6,6 +6,9 @@ from exceptions import TimeoutException
 DEFAULT_RETRY_TIMEOUT = 60
 DEFAULT_QUERIES_PER_SECOND = 10
 
+TOMTOM_403_RATE_LIMIT_HEADER = 'Account Over Queries Per Second Limit'
+TOMTOM_DETAIL_HEADER = 'X-Error-Detail-Header'
+
 
 def qps_retry(original_function=None, **options):
     """ Query Per Second retry decorator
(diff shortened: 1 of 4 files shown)


ethervoid:
@antoniocarlon changes added pls could you review it again?

server/lib/python/cartodb_services/cartodb_services/tools/qps.py
```diff
@@ -46,6 +48,8 @@ def call(self, fn, *args, **kwargs):
                 response = getattr(e, 'response', None)
                 if response is not None:
                     if self._provider is not None and self._provider == 'tomtom' and (response.status_code == 403):
+                        if response.headers['X-Error-Detail-Header'] != TOMTOM_403_RATE_LIMIT_HEADER:
```
antoniocarlon: We could define `X-Error-Detail-Header` as a constant and get it using `get` to avoid errors

ethervoid: done

antoniocarlon: We can get the value using `get` to avoid errors:
`if response.headers.get('X-Error-Detail-Header') != TOMTOM_403_RATE_LIMIT_HEADER:`

ethervoid: I don't see the point xD

antoniocarlon: doing `response.headers['X-Error-Detail-Header']`, if `response.headers` doesn't contain the key `'X-Error-Detail-Header'` it will fail with `KeyError`. We can avoid this error using `get`

ethervoid: Ohh I see, done thx

ethervoid:
Tested in staging and seems to work fine

PR #539, opened 2018-12-24
S1891 bulk geocoder no street address

javitonino:
Fix for https://github.com/CartoDB/support/issues/1891


gonzaloriestra:
:heavy_check_mark: Acceptance OK

Checked that it fails with the provided CSV and that it works after applying the patch. Tested for regular and bulk process (<100 and >100 rows).

PR #543, opened 2019-02-19
Fix bulk geocoding soft limit

rafatower:
Fix the case where there's not enough quota to geocode a table but the
soft_limit is set to true.

The function `cdb_enough_quota` accounts for the `soft_limit` flag, as
well as for the remaining quota and the rows to be geocoded.

### DIFF
--- a/client/sql/21_bulk_geocoding_functions.sql
+++ b/client/sql/21_bulk_geocoding_functions.sql
@@ -49,7 +49,7 @@ BEGIN
   RAISE DEBUG 'cdb_bulk_geocode_street_point --> query_row_count: %; query: %; country: %; state: %; city: %; street: %',
       query_row_count, query, country_column, state_column, city_column, street_column;
   SELECT cdb_dataservices_client.cdb_enough_quota('hires_geocoder', query_row_count) INTO enough_quota;
-  IF remaining_quota < query_row_count THEN
+  IF NOT enough_quota THEN
     RAISE EXCEPTION 'Remaining quota: %. Estimated cost: %', remaining_quota, query_row_count;
   END IF;
 
--- a/client/sql/21_bulk_geocoding_functions.sql
+++ b/client/sql/21_bulk_geocoding_functions.sql
@@ -49,7 +49,7 @@ BEGIN
   RAISE DEBUG 'cdb_bulk_geocode_street_point --> query_row_count: %; query: %; country: %; state: %; city: %; street: %',
       query_row_count, query, country_column, state_column, city_column, street_column;
   SELECT cdb_dataservices_client.cdb_enough_quota('hires_geocoder', query_row_count) INTO enough_quota;
-  IF NOT enough_quota THEN
+  IF enough_quota IS NULL OR NOT enough_quota THEN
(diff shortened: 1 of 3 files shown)


client/sql/21_bulk_geocoding_functions.sql
```diff
@@ -49,7 +49,7 @@ BEGIN
   RAISE DEBUG 'cdb_bulk_geocode_street_point --> query_row_count: %; query: %; country: %; state: %; city: %; street: %',
       query_row_count, query, country_column, state_column, city_column, street_column;
   SELECT cdb_dataservices_client.cdb_enough_quota('hires_geocoder', query_row_count) INTO enough_quota;
-  IF remaining_quota < query_row_count THEN
+  IF NOT enough_quota THEN
```
juanignaciosl: Not sure about the change. Wondering how come I used the comparison instead of just `enough_quota` it looks like I changed it [here](https://github.com/CartoDB/dataservices-api/commit/531ad281#diff-b6b480d3a8e84ab4b9b8e1e97f9afa19R2028). It certainly looks intentional.

Looking inside `cdb_enough_quota` it finally returns `params.soft_limit OR ((params.used_quota + input_size) <= params.monthly_quota`. The right side of the `OR` is equivalent to the current check, which didn't have `params.soft_limit`, so it looks like I intentionally removed it.

Maybe it was an attempt to avoid users from being charged for huge numbers? TBH, I can't tell. Give it a thought, maybe you're closer to this now and can come up with something interesting. The commit included some tests, hopefully they would raise an alarm if this were wrong.

The check previously looked this way: `IF enough_quota IS NOT NULL AND NOT enough_quota THEN`, maybe adding the `IS NOT NULL` would be safer, up to you.

rafatower: > so it looks like I intentionally removed it

There must be a ticket or something to keep track of that decision. Otherwise we're gonna be changing it forever.

I implemented it because of a RT complaint (not honoring current documentation BTW)

> Give it a thought, maybe you're closer to this now and can come up with something interesting

yes, implement two different thresholds. let's call them "soft limit" and "hard limit":
- "soft limit": warn the user about it being crossed
- "hard limit": stop the user from consuming any more quota, to avoid unexpected expenses.

but that's way beyond RT.

Who should make that call? cc/ @inigomedina

juanignaciosl: > There must be a ticket or something to keep track of that decision. Otherwise we're gonna be changing it forever.

It was part of the development of the optimal batch size (check the commit comment). I agree upon changing it if there's not a good reason not to do so. Maybe I was just wrong with that, I just wanted to try to find the reason for the change.

jgoizueta:
I have finally changed the `IF NOT enough_quota` by a `IF enough_quota IS NULL OR NOT enough_quota` to prevent the possibility of a bad configuration producing a NULL value of enough_quota which, given how SQL handles NULLs, would skip the raise and proceed as if quota was enough.

Note that if a service is simply not configured we'll get an exception from `cdb_enough_quota`. But some evil configuration could produce a NULL. I (and @rafatower ) think the right place to handle that case would be the function(s) that read the configuration, but I've taken this decision here to expedite the deployment of this fix.

PR #551, opened 2019-03-13
Fix altered public method of TomTomGeocoder

jgoizueta:
In #456 the public method geocode_meta return value was changed (to a tuple). The intention was to be consumed only from the same class.


jgoizueta:
I've tested this in staging and is working.

PR #552, opened 2019-03-13
Release python-lib 0.21.4

jgoizueta:
Fix TomTom bulk geocoder bug (#551)


jgoizueta:
This seems to be already in master  ¯\\\_(ツ)\_/¯

PR #572, opened 2020-01-17
[leapfrog] First implementation of the true Mapbox isochrones

antoniocarlon:
First implementation of the true Mapbox isochrones.

It keeps the previous `mapbox` provider for the existing isochrones and adds a new `mapbox_iso` provider for the true Mapbox isochrones. The idea is to allow some users to keep the old on (default) while creating a new one for other users, mostly to ensure that the Mapbox pricing for the new isochrones suit us.

Related to https://github.com/CartoDB/cartodb-platform/issues/6141


server/lib/python/cartodb_services/cartodb_services/mapbox/true_isolines.py
```diff
@@ -0,0 +1,154 @@
...
+PROFILE_WALKING = 'walking'
+DEFAULT_PROFILE = PROFILE_DRIVING
+
+MAX_TIME_RANGE = 60 * 60  # The maximum time that can be specified is 60 minutes.
+                          # https://docs.mapbox.com/api/navigation/#retrieve-isochrones-around-a-location
+
+MAX_SPEEDS = {
+    PROFILE_WALKING: 3.3333333,  # In m/s, assuming 12km/h walking speed
+    PROFILE_CYCLING: 16.67,  # In m/s, assuming 60km/h max speed
+    PROFILE_DRIVING: 41.67  # In m/s, assuming 140km/h max speed
```
rafatower: vamos, neno! are you sure?
```
140 km / h = 140000 m / 3600 s = 38.88 m/s
```

server/lib/python/cartodb_services/test/metrics/test_config.py
```diff
@@ -196,6 +196,8 @@ def test_should_return_user_config_for_isolines(self):
                 assert isolines_config.service_type is 'mapzen_isolines'
             elif isolines_provider is 'mapbox':
                 assert isolines_config.service_type is 'mapbox_isolines'
+            elif isolines_provider is 'mapbox_iso':
+                assert isolines_config.service_type is 'mapbox_iso_isolines'
```
rafatower: I think this will affect limits, billing, etc and would make it unnecessarily complicated. Did you consider just replacing the old method with the new one?

server/extension/sql/80_isolines_helper.sql
```diff
@@ -188,6 +188,70 @@ RETURNS SETOF cdb_dataservices_server.isoline AS $$
...
+    service_manager.quota_service.increment_isolines_service_use(len(isolines))
+    return result
+  except BaseException as e:
+    import sys
+    service_manager.quota_service.increment_failed_service_use()
+    service_manager.logger.error('Error trying to get Mapbox true isolines', sys.exc_info(), data={"username": username, "orgname": orgname})
+    raise Exception('Error trying to get Mapbox true isolines')
+  finally:
+    service_manager.quota_service.increment_total_service_use()
+$$ LANGUAGE plpythonu SECURITY DEFINER STABLE PARALLEL RESTRICTED;
```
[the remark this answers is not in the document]
antoniocarlon: :thinking: these functions have been defined with `SECURITY DEFINER` since long time ago

server/lib/python/cartodb_services/cartodb_services/mapbox/true_isolines.py
antoniocarlon: Fixed! :+1:

server/lib/python/cartodb_services/test/metrics/test_config.py
antoniocarlon: My idea here was to separate the old implementation from the new implementation so we can still offer the old version for existing users using Mapbox as the isolines provider without breaking their workflow and ensuring that the Mapbox true isolines fit our needs in terms of pricing, but I'll add other stakeholders to the look to ensure that this approach is correct

antoniocarlon: According to the discussion here https://github.com/CartoDB/cartodb-platform/issues/6141 I have finally replaced the old mapbox implementation with the new one :+1:

antoniocarlon:
@rafatower second round?

antoniocarlon:
I have removed the generated files to ease review. I'll add them when I get green light :+1:

server/lib/python/cartodb_services/cartodb_services/mapbox/isolines.py
```diff
@@ -1,171 +1,142 @@
...
 }
 
-DEFAULT_NUM_ANGLES = 24
-DEFAULT_MAX_ITERS = 5
-DEFAULT_TOLERANCE = 0.1
-
-MATRIX_NUM_ANGLES = DEFAULT_NUM_ANGLES
-MATRIX_MAX_ITERS = DEFAULT_MAX_ITERS
-MATRIX_TOLERANCE = DEFAULT_TOLERANCE
+VALID_PROFILES = [PROFILE_DRIVING,
```
rafatower: Minor thing: how about `VALID_PROFILES = (...)` when you don't intend to push/pop things from the array?

server/lib/python/cartodb_services/cartodb_services/metrics/config.py
```diff
@@ -676,10 +671,9 @@ def _get_mapbox_config(self):
             raise ConfigException('Mapbox configuration missing')
 
         mapbox_conf = json.loads(mapbox_conf_json)
-        self._mapbox_matrix_api_keys = mapbox_conf['matrix']['api_keys']
-        self._mapbox_matrix_quota = mapbox_conf['matrix']['monthly_quota']
-        self._mapbox_matrix_service_params = mapbox_conf['matrix'].get('service', {})
-        self._mapbox_isochrones_service_params = mapbox_conf.get('isochrones', {}).get('service', {})
+        self._mapbox_isolines_api_keys = mapbox_conf['isolines']['api_keys']
```
rafatower: Out of curiosity: do you need to populate new keys? or are they the same? If they're new, you need to populate them prior to deploying in order not to disrupt the service.

If they are the zame, I'd re-use them, adding an alias and/or a note in the code, so that you can spare a that operation.

antoniocarlon: The keys are the same and I had prepared a PR on platform to address this issue, but I'll follow your guide here and keep the old parameters to avoid breaking anything :+1:

antoniocarlon:
@rafatower I have addressed your comments in case you want to give it another review :+1:

rafatower:
I trust you and your tests, so please go ahead :muscle:

PR #578, opened 2020-02-12
Added Mapbox true isolines

antoniocarlon:
Added Mapbox true isolines


antoniocarlon:
The PG10 tests are green and after some local testing it seems that the PG12 tests are failing because of the `plpythonu` vs `plpython3u` extensions.

This is something that we need to take a deeper look into when we migrate to PG12, but right now I'm merging and deploying.

PR #584, opened 2020-02-27
[PG12] Adapt project to PostgreSQL 12 and PostGIS 3

manmorjim:
This fixes https://github.com/CartoDB/cartodb-platform/issues/6237

### DIFF
--- a/client/sql/05_utils.sql
+++ b/client/sql/05_utils.sql
@@ -18,3 +18,16 @@ $func$ LANGUAGE plpgsql;
 CREATE OR REPLACE FUNCTION cdb_dataservices_client.cdb_jsonb_array_casttext(jsonb) RETURNS text[] AS $f$
     SELECT array_agg(x) || ARRAY[]::text[] FROM jsonb_array_elements_text($1) t(x);
 $f$ LANGUAGE sql IMMUTABLE;
+
+
+-- PG12_DEPRECATED
+-- Create geomval if it doesn't exist (in postgis 3+ it only exists in postgis_raster)
+DO $$
+BEGIN
+    IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'geomval') THEN
+        CREATE TYPE geomval AS (
+            geom geometry,
+            val double precision
+        );
+    END IF;
+END$$;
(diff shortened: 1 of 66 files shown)


client/sql/05_utils.sql
```diff
@@ -17,4 +17,17 @@ $func$ LANGUAGE plpgsql;
...
-$f$ LANGUAGE sql IMMUTABLE;
\ No newline at end of file
+$f$ LANGUAGE sql IMMUTABLE;
+
+
+-- PG12_DEPRECATED
+-- Create geomval if it doesn't exist (in postgis 3+ it only exists in postgis_raster)
+DO $$
+BEGIN
+    IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'geomval') THEN
```
Algunenano: What schema is used to create this type? Is it in public? Would it be possible to create it in another schema so it doesn't conflict with postgis_raster?
I don't know how we are referencing it, if we use public.geomval then we do have to use public.

Algunenano: Same for the server

server/extension/test/sql/366_empty_table_test.sql
```diff
@@ -37,7 +42,7 @@ GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA cdb_observatory TO geocoder_api;
 SELECT * FROM cdb_dataservices_server.OBS_GetData(
     'foo',
     NULL,
-    '{"(0103000020E61000000100000005000000010000E0F67F52C096D88AE6B25F4440010000E0238052C0BF6D8A1A8D5D4440010000D0DA7E52C05F03F3CC265D444001000020F47E52C0F2DD78AB5D5F4440010000E0F67F52C096D88AE6B25F4440,1)"}'::public._geomval,
+    '{"(0103000020E61000000100000005000000010000E0F67F52C096D88AE6B25F4440010000E0238052C0BF6D8A1A8D5D4440010000D0DA7E52C05F03F3CC265D444001000020F47E52C0F2DD78AB5D5F4440010000E0F67F52C096D88AE6B25F4440,1)"}'::_geomval,
```
Algunenano: We should confirm that nowhere else, at least in our projects, we use public.geomval.

PR #590, opened 2020-03-10
Fixes client release 0.29.0

manmorjim:


Algunenano:
Why are we editing old releases? We shouldn't be installing those in PG12+? Did we get any error to justify this change?

skgsergio:
Well maybe old releases editing is not required, but adding the replacement of variables to them is.

PR #599, opened 2020-05-19
Enabling HIPAA, fix bulk geocode output

antoniocarlon:
- Enabling HIPAA
 - Fix bulk geocode output


antoniocarlon:
@xavipereztr, are our Geocod.io API keys they HIPAA enabled?

With this change we would need them to be HIPAA enabled or the geocoding with Geocod.io from our cloud will fail with ' Geocodio authorization error`

antoniocarlon:
*Note:*  The tests are failing because our current API key is not HIPAA compliant

xavipereztr:
Hi,

The API Key for the HIPAA compliant end-point is: 44b3537089b3e9f75304c7f5efcb9fff09c733e

PR #609, opened 2020-11-17
Improve error reporting

Algunenano:
Related to ch119936

Improves how errors are reported
- Stop calling `ServiceException` with a status code. The status code is a int, so it will crash (`TypeError: __str__ returned non-string (type int)`).
- Routing: When receiving an unknown error, report it to the caller instead of swallowing. For example: `ERROR:  cdb_dataservices_client._cdb_route_point_to_point(7): [dataservices_db] REMOTE ERROR: spiexceptions.ExternalRoutineException: Exception: Error trying to calculate TomTom routing: Unexpected response (500): {"formatVersion":"0.0.12","error":{"description":"Engine error while executing route request: STOPPED"},"detailedError":{"message":"Engine error while executing route request: STOPPED","code":"INTERNAL_ERROR"}}`. Beforehand it stopped after ` Error trying to calculate TomTom routing` so there was no way of knowing what was going on (in this case a 500 error from Tomtom reporting an internal issue). I would rather be verbose here than super terse as before.

Needs new releases of the server (both python and sql).


antoniocarlon (approved):
LGTM! :rocket:

PR #614, opened 2021-01-26
[ch127801] Country conversion to ISO code in TomTom bulk geocoding

Shylpx:
### Resources

- [Clubhouse story](https://app.clubhouse.io/cartoteam/story/127801/fix-countryset-for-tomtom-geocoding)

### Context

- The TomTom requests require the country param (`countrySet`) as a ISO code, according to the TomTom documentation:
Screenshot from 2021-01-26 15-19-28
- There are two different methods for geocoding using TomTom provider:
  - [_cdb_tomtom_geocode_street_point](https://github.com/CartoDB/dataservices-api/blob/d3720d6953e9435d43972c798ce1ba3e3c56f6e0/server/extension/cdb_dataservices_server--0.39.2.sql#L2270): Single address geocoding.
  - [_cdb_bulk_tomtom_geocode_street_point](https://github.com/CartoDB/dataservices-api/blob/d3720d6953e9435d43972c798ce1ba3e3c56f6e0/server/extension/cdb_dataservices_server--0.39.2.sql#L2518): Bulk address geocoding.
- So, when a country name is passed as a `country` param to those methods, it should be converted to restrict the results to that country. There is already a conversion in the first one using the `country_to_iso3` method, but while bulk geocoding, there is no conversion being performed.

### Changes

- Add a conversion from country name to ISO code in the TomTom bulk geocoder.


server/lib/python/cartodb_services/cartodb_services/tomtom/bulk_geocoder.py
```diff
@@ -35,6 +36,9 @@ def _serial_geocode(self, searches):
         results = []
         for search in searches:
             (search_id, address, city, state, country) = search
+
+            country = country_to_iso3(country) or country
```
Algunenano, line 40: You are only changing the serial geocoder of the bulk geocoder (I think it's used when the array passed to the bulk geocoder is small). I think this should be done in a common path.

Algunenano (changes requested):
PG12 tests seem broken. I think Infra did some changes to the base images so that might be way. It should be fixed before merging this.

server/lib/python/cartodb_services/cartodb_services/tomtom/bulk_geocoder.py
Shylpx: The common path is the `_request_uri` method in the `TomTomGeocoder`. Since it is also being used by the `cdb_dataservices_server._cdb_tomtom_geocode_street_point`, I removed the existent conversion from that method to avoid a double conversion.

Shylpx:
I have asked infra in order to fix the PG12 build before merging.

Algunenano (approved):
Looks fine to me. Nice!

PR #628, opened 2022-09-01
Fix parameter for new HERE Isolines API

moicalcob:
After updating to the HERE Isolines V8 API, some customers are experiencing errors while trying to calculate isolines by destination instead of origin. This PR fixes this strange behaviour.


Shylpx (approved):
:rocket:

PR #635, opened 2023-08-17
Add mode_feature parameter to HERE V8 isolines

moicalcob:
Add parameter `mode_feature` that was lost due to the migration from HERE API v7 to v8 a few months ago.

```
dataservices_1                | DEBUG:urllib3.connectionpool:https://isoline.router.hereapi.com:443 "GET /v8/isolines?origin=40.420280%2C-3.705680&range%5Bvalues%5D=1800&apikey=xxxx&routingmode=short&optimizefor=quality&avoid%5Bfeatures%5D=tollroad&transportmode=car&range%5Btype%5D=time HTTP/1.1" 200 None
```

### DIFF
--- a/server/lib/python/cartodb_services/cartodb_services/here/routing.py
+++ b/server/lib/python/cartodb_services/cartodb_services/here/routing.py
@@ -180,7 +180,8 @@ class HereMapsRoutingIsolineV8(Traceable):
         'departure',
         'arrival',
         'maxpoints',
-        'quality'
+        'quality',
+        'mode_feature'
     ]
 
     def __init__(self, apikey, logger, service_params=None):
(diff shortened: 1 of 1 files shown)


Mmoncadaisla (approved):
Thank you for taking care of this one @moicalcob this is greatly appreciated ❤️ 

As discussed together, I'm Ok with these changes ✅ specially given that this functionality is not documented in the Data Services API documentation, but noting that if other usage besides strictly avoiding tollroads was required (which is not the case), we should probably revisit this and make sure we:

- Map every option for the `mode[feature]` parameter in v7 to the correspondent value in the `avoid[feature]` parameter in v8 (e.g: "motorway" to "controlledAccessHighway")
- Consider raising a `ValueError` or similar exception when the CARTO `cdb_isochrone` function leverages a value for `mode_feature_weight` different from `-3` (strictly avoid), as this has changed in HERE API v8 and only strict is supported, or at least document this behavior.

>On a very small side note that can be completely ignored, it could have been more aligned with the current approach to treat the `mode_feature` parameter parsing as part of the `__get_mode_params` method instead. 

Thanks again for your help!

